PR #4, opened 2019-10-17
Run dev chain with docker

kaichaosun:
Related https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/283

TODO:

- [ ] Run multi node with Docker.


gabreal:
The [CI builds](https://github.com/paritytech/substrate/blob/master/.gitlab-ci.yml#L229) substrate and publishes the binary on S3 and creates a [docker image](https://hub.docker.com/r/parity/substrate/tags) from it. Apart from that there is a contributed [Dockerfile](https://github.com/paritytech/substrate/blob/master/Dockerfile) for building and creating a docker image.

Can they possible be extended to server your purpose?

kaichaosun:
@gabreal I noticed this dockerfile, it has two stages, I actually need the first stage. This is [the attemp](https://github.com/substrate-developer-hub/substrate-node-template/pull/4/commits/79a910b905ed6ed2e7998204c5fa4cc464f2e7b2) that I try to use this dockerfile by modifying it locally.
Then I find `parity/rust:nightly`, it also works in this case. In this option, I don't need to change the build pipeline to release another base image for node-template in Substrate repo. 
I do think it's valuable to have a standalone image specific for runtime developers, just substrate repo seems not the good place to do the job. Probably the [scrips repo](https://github.com/paritytech/scripts)?

gabreal:
Do I understand correctly that your plan is to have a docker image to complile rust code?

I that case i think it makes sense to use the `rust-builder` image that is also used in the substrate ci itself (from https://github.com/paritytech/scripts/blob/master/dockerfiles/rust-builder/Dockerfile to https://hub.docker.com/r/parity/rust-builder/).

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+  dev:
+    image: parity/rust:nightly
+    working_dir: /substrate-node-template
+    ports:
+      - "9944:9944"
+    volumes:
+      - ".:/substrate-node-template"
+      - "cargo-registry-cache:/root/.cargo/registry"
+      - "cargo-git-cache:/root/.cargo/git"
+    command: bash -c "cargo run --release -- --dev --ws-external"
```
gabreal: afaik, bash is not needed here, is it?

kaichaosun: You are right.

JoshOrndorff:
Looks awesome @kaichaosun. I would love to know more about these docker images and how they could be used at SFBW. Would you mind showing me how this works on a zoom call?

kaichaosun:
Sure, @JoshOrndorff PMed you.

shawntabrizi:
Can we get Windows compatible instructions?

I think that is the main use case for docker right? (since linux and macos have pretty easy setup)

kaichaosun:
@shawntabrizi Good point, will test the script in Windows.

shawntabrizi:
Non-bash script right? If they have a linux subsystem, then the original getsubstrate script will work AFAIK?

kaichaosun:
The dependencies that installed by getsubstrate script are preinstalled in the docker image, should be working from what I know.

jimmychu0807:
@kaichaosun can you also send an zoom invite to me if you are going to demo it to Joshy

riusricardo:
@kaichaosun can you also include me :) I'm looking a lot into docker to cross-compile and this will be helpful.

kaichaosun:
Tested with ubuntu 18.04.3 LTS and mac os, works fine. 
I don't have a windows pro which supports docker, shall I merge it for now and seek help from our community? @shawntabrizi

shawntabrizi:
@kaichaosun keep these changes on a local fork of the Substrate Node Template. We can share that way for now.

I dont want to merge to the main repo because I don't want to imply that we will for sure support this long term

kaichaosun:
Close it for now, will open another pr once solve the size of docker image and other concerns.

PR #12, opened 2019-12-03
Running node-template with substrate v2

jimmychu0807:
While upgrading my offchain worker demo to run in substrate v2, I think it is the best to copy paste my work on top of a node-template that run v2, thus this PR. Compiled against [this substrate version](https://github.com/paritytech/substrate/commit/368318c9b11c07596cde5dc37220258cfa7dfa76).

I am basically copying from `SUBSTRATE-PROJECT/bin/node-template` and updating `Cargo.toml` to points to remote substrate git repo.


shawntabrizi:
@jimmychu0807 FYI there is a script you can run which quickly creates a "working node template" with all the right cargo dependencies for Substrate:
https://github.com/paritytech/substrate/blob/master/.maintain/node-template-release.sh

This script is actually automatically run everytime a new commit to Substrate master happens:

http://releases.parity.io/substrate/x86_64-debian:stretch/2.0.0-<9-digit-commit>/substrate/substrate-node-template.tar.gz


Just update <9-digit-commit> to the first 9 digits of a commit hash on Substrate master.

**RE: This PR**

I think we should wait for a v2.0 branch before making updates to this repository. It will just cause pain to our currently working tutorials.

JoshOrndorff:
>  FYI there is a script you can run...

Should that info live in the readme or somewhere easier to find for future contributors?

shawntabrizi:
IMO not really. They should use our pre-packaged github :)

I don't see when we would encourage this to a user.

PR #16, opened 2019-12-27
readme update on generating your own node-template

jimmychu0807:
I think this is a good tip and want developers to know about. So they can choose to diy their node-template in case we don't update the node-template soon enough ;)


README.md
```diff
@@ -74,3 +74,27 @@ cargo run -- \
...
+
+# Switch to a branch or commit of the Substrate repo you want your node-template based on
+# git checkout <branch/sha1>
+
+cd substrate/.maintain
+
+# Run the helper script to generate a node template.
+# This script compiles Substrate and takes a while to complete. It takes a relative file path
+#   from the current dir. to output the compressed template file.
+./node-template-release.sh ../../node-template.tar.gz
```
riusricardo: It can also be done from the main substrate directory without the need to cd into `.maintain`
```
.maintain/node-template-release.sh node-template.tar.gz
```

README.md
```diff
@@ -74,3 +74,27 @@ cargo run -- \
 ```
 
 Additional CLI usage options are available and may be shown by running `cargo run -- --help`.
+
+### Generate Your Own Substrate Node Template
```
riusricardo: I'm not sure if we should encourage this. The problem will be that master could introduce breaking changes as it evolves. Supporting `unstable` versions should remain out of scope for the dev hub.

jimmychu0807: One reason I find this helpful is when I want to generate a node template that is compatible to a specific old version of substrate. I am readers are already hacking on Substrate (since they are willing to clone the main Substrate code repo), they maybe playing with a certain branch, or certain old version. This cmd enables them to generate a template that they can build their runtime on. 

We can add a note saying this is an **Advanced Feature** etc, to let read know to use it at their own risk.

JoshOrndorff: I prefer adding this info. IMO it's better to be documented than undocumented.

I'm also fine with adding language calling it advanced, or saying users will get faster and more thorough support if they stick with the releases here.

riusricardo: Agree with both. I forgot to reply here. It is fine if we add an advanced title/warning.

jimmychu0807: updated with both of your note. Thanks!

PR #22, opened 2020-01-29
Directories Restructuring

jimmychu0807:
@JoshOrndorff  @riusricardo  @joepetrowski 

While I am working on the offchain worker project, I am figuring what's the best way to store node, runtime, and the price-fetch pallet.  I realized node to runtime is a 1-to-1 relationship. Runtime to pallet is a 1-to-many relationship. The reason we put `runtime` into a separate folder is because it requires a different compilation method from `node`. We need a `build.rs` to compile runtime to both native and wasm.

I propose restructuring the node-template to the following (with @gnunicorn dir structure suggestion):
Also taking @riusricardo suggestion of putting mock and tests in a single file `tests.rs`.

```
node
  // this is the node, main entry point. 
  // folder renamed from `src` to `node` to make its meaning clearer
  L main.rs
  L ...
  L runtime
    // this is our runtime
    L src
      L lib.rs
    L build.rs
    L Cargo.toml
pallets
  L template
    // this is our `template` pallet 
    L src
      L lib.rs
      L tests.rs
    L Cargo.toml
Cargo.toml
```

The advantage are:
  - it is clear to dev that the node, runtime, and pallets are "separated"
  - You can easily extend this structure to write multiple pallets. 
  - We are also introducing the structure of `substrate` main codebase to dev. They can 
familiarize to this structure early on.
  - With this structure, I think we can phase out `substrate-pallet-template`. Because whatever inside the `/pallets` is already a good pallet development example. With its standalone code, mock runtime, test cases, and its `Cargo.toml`, one just need to take this folder out and it is a good pallet to upload to git / cargo crate.

I am aware this will cause some documentation change. We can do this when we have time to update the docs, and merge the PR when the docs are ready. But I feel this is a better way to structure node template in the long run.


jimmychu0807:
Joe:
> I would be interested in what Ben or Tomek think about this - I agree with your logic Jimmy but I'm also not in the codebase enough to have a strong voice on this

@gnunicorn  @tomusdrw Would like to hear if you have any feedback on this PR of directory restructuring for substrate-node-template. 

Reiterating on the purpose, substrate-node-template is for new developers to easily setup and run a node. It is a bonus if they can also pick up some best practices / general design pattern for substrate runtime development.

tomusdrw:
I like the idea. Separating from `runtime` emphasises that `pallets` can be reused in different runtimes (possibly with different generic parameters). If we suggest writing `pallets` in way that does not depend on runtime primitives (i.e. generic), it would make it easy for devs to publish their work in the future.

Currently putting `pallets` inside `runtime` kind of endorses writing non-generic code - obviously it's just much simpler, but can lead to more spaghetti.

Now I thought of it I think we can actually suggest a mix of the two approaches:
1. Put the runtime-specific stuff within `runtime` (as currently)
2. Put potentially re-usable `pallets` in it's own dir, but make sure that: it doesn't depend directly on other `pallets` from the same dir AND it by no means it depends on anything from `runtime`.

Cargo.toml
```diff
@@ -1,5 +1,8 @@
-[profile.release]
-panic = 'unwind'
+[workspace]
```
tomusdrw: I don't think workspaces makes sense here. We are building a single binary anyway and that binary is probably referencing everything that is in the same repo. But maybe there are some other consequences I'm not aware of.

gnunicorn:
I like this, too. Good stab at it!

In particular I like that this enforces a stronger separation as pallets being independent crates. This makes the template a good default to build your own chain, but also to create a pallet-develop-environment–very cool!

I have to general questions, I would like to discuss regarding the structure (I do not have answers to them yet):
1. we could also rename `src` to `client` or `node` (you can specify the path in the `Cargo.toml`) to make it more clear what this is and how it relates. In particular if you care about implementing a custom pallet, that repo is just the demo-implementation and – unlike the current naming suggests – the main source entry point.
2. In the same light and idea of clearing out what it the most important thing you will be working on, I wonder if we should move `runtime` _into_ `src`/`node`/`client`. For once because it _runs inside_ the node and secondly, the most changes will be done in `pallets`, so having two other top-level folders, one of them not being touched ever (`src`) seems to create an uneven playing field.  


:thinking: going from the 2nd, and the pallets being the most important thing, I wonder why not move them to the top-level and put everything else under `demo`?

Cargo.toml
gnunicorn: `cargo test` wouldn't necessary test local crates that aren't referenced here, would it? It would only test the main crate. I think a workspace is good practice.

jimmychu0807: Thinking about it, I may just expose the `template-pallet` out. So people can run tests on a particular pallet (`cargo test -p pallet-*`). But there is no need to expose `runtime` as a package by itself as of now.

jimmychu0807:
@gnunicorn I love the two suggestions! Will rework on them.

jimmychu0807:
@riusricardo  @JoshOrndorff 

The restructuring is done with team suggestions. The latest structure is outlined in the topmost comment.

The only thing is we now need a migration plan to update docs and any other dependencies so this PR can be merged.

riusricardo:
Looks good :+1: 
We need to create a list of the affected tutorial/docs and then the PRs before merging this.

JoshOrndorff:
I like this design a lot. It's similar to what I proposed in https://github.com/paritytech/substrate/pull/3769

My one hesitation is that merging this PR here causes us to diverge from the `node-template` that is in the main Substrate repo. One option would be to make these changes upstream in Substrate repo first. Then our process for updating docs etc is to add TODO_RELEASE markers to our code, and then cut a new node-template by copying from Substrate repo when 2.0 is released.

jimmychu0807:
Joshy, thanks for your reply. First, you reminded me that substrate-node-template is not a standalone package, but generated from substrate repo. I need to modify that. Second, thanks for sharing the whole historical context / discussion. 

1. I will still do a Substrate PR later today/tmr. Should I @ anyone in that domain for PR review?
2. In the discussion thread paritytech/substrate#3769, 

> can't we test this idea on customers using the package/templates we have set up on Developer Hub?
>
> I am not saying to throw the change away, I am simply saying to test with users how it affects their onboarding experience.

On a side note, Do you know how can we "test this idea on customers/users" now? Do we have contact with some testers asking them to review our docs or try out our templates?

jimmychu0807:
@JoshOrndorff  @riusricardo 

This is the final structure of how node-template looks like now: https://github.com/paritytech/substrate/pull/4811#issuecomment-582273073

JoshOrndorff:
This went in upstream in paritytech/substrate#4811 and will be included in the next release of node template. Closing this PR.

PR #25, opened 2020-03-05
Update node template to alpha.3

JoshOrndorff:
This is a standard update of the node template to a newer upstream version.

One weird thing about it is that although I checkout out `alpha.3` and then ran the release script, many of the dependencies in `Cargo.toml` still point to `alpha.2`. @gnunicorn Do you know anything about that. Is it expected? Should I manually update them to be `alpha.3`?

I'd like to use this as a target for updating the devhub tutorials, but want to be sure the dependencies are correct first.


jimmychu0807:
@gnunicorn  In fact, using current Substrate master (which build on top of `v2.0.0-alpha.3`), if I search for `alpha.2` string, it is found in various `Cargo.toml` (more than 1000 matches). Can I assume all `alpha.2` to be replaced by `alpha.3`?

JoshOrndorff:
image

So I'm gonna update all the alpha.2 to alpha.3 manually for this one release. It seems it will be sorted in the future.

riusricardo:
Does this branch works with polkadot.js.org/apps? I was doing the same yesterday but it fails with a Multisigner error in apps.

JoshOrndorff:
You can get past the multisig error by adding these types
```json
{
  "Address": "AccountId",
  "LookupSource": "AccountId"
}
```
The reason they are needed is explained here: https://polkadot.js.org/api/start/types.extend.html#impact-on-extrinsics
The change went into main Substrate repo here: https://github.com/paritytech/substrate/pull/5025

Although now that I'm past that error my extrinsics are still failing with `Invalid`. Not a super helpful error message :-/

JoshOrndorff:
With help from @jacogr I discovered that the problem with transfers was block weight limitations. The maximum block weight is set to 1 million [source]() and a transfer weighs 1 million [source](). So after a timestamp inherent is in the block there isn't room left for even a single balance transfer.

I've confirmed that raising the max block weight solves the problem, so now we have two questions:
1. None of the numbers I mentioned above has changed in the last 8 months (since [this PR](https://github.com/paritytech/substrate/pull/3157)) so why are we just having this problem now?
2. How do we fix it? If we only fix it upstream in Substrate, then we can't start using this alpha.3 release. If we fix it here, our node template has diverged from the one upstream in Substrate.

JoshOrndorff:
My preference is to increase the weight limit manually here. This makes our node template slightly different than the one in Substrate master at alpha.3, but I'm okay with that, and it won't be a long-term divergence.

I'd like to merge this and tag the alpha.3 release.

jacogr:
So the API does allow for node-specific types based on spec_name & version. So issues like https://github.com/substrate-developer-hub/substrate-node-template/pull/25#issuecomment-595789826 can be transparently resolved on both 

- API (prefer not to keeping to Polkadot/Substrate master only, https://github.com/polkadot-js/api/blob/master/packages/types/src/known/overrides.ts)
- UI (generally preferred, https://github.com/polkadot-js/apps/tree/master/packages/apps-config)

However, it is not something I myself can maintain, I have enough to do and never run the templates either, I have enough branches to test through on any given day. So feel free to make a PR and maintain there for easy getting started.

shawntabrizi:
> You can get past the multisig error by adding these types
> 
> ```json
> {
>   "Address": "AccountId",
>   "LookupSource": "AccountId"
> }
> ```
> 
> The reason they are needed is explained here: https://polkadot.js.org/api/start/types.extend.html#impact-on-extrinsics
> The change went into main Substrate repo here: [paritytech/substrate#5025](https://github.com/paritytech/substrate/pull/5025)
> 
> With those types most extrinsics work. However balance transfers still fail with `Invalid`. Not a super helpful error message :-/

I am going to revert this change. Sorry about any issues

riusricardo:
Thanks @shawntabrizi, this will make everyone's life easier. I'm not a fan of explaining the custom types in all of the tutorials.
We discussed the possibility of adding back indices. Doing it would require to revert you PR, however your argument to remove it was also valid.

JoshOrndorff:
Shawn's remove-indices-from-template change was made for good reasons. I actually like the node template better without indices for exactly the reasons he described.

@jacogr To make a PR to Apps as described in your link.
* Is this the spec string we would use https://github.com/substrate-developer-hub/substrate-node-template/blob/master/runtime/src/lib.rs#L95 ?
* And adda file (plus minor change to `index.ts` here https://github.com/polkadot-js/apps/tree/master/packages/apps-config/src/api/spec ?

jacogr:
@JoshOrndorff 100% 

With or without `-`'s in the name (Those get filtered out since some nodes started swapping to not using them... and then creating chaos with old and new. But probably best to include them to be 100% that is is what it is, need to revert some atm to their "as-in-code" names there)

JoshOrndorff:
No that https://github.com/polkadot-js/apps/pull/2364#event-3115424903 is merged, Apps will work with the node template with no custom types added by the user. It seems that was the last potential hold-up to merging this.

jacogr:
Does it work? ;)

riusricardo:
It does @jacogr.
Good job @JoshOrndorff :+1:

PR #29, opened 2020-04-01
Delete outdated build.rs

JoshOrndorff:
This PR deletes a single `build.rs` file from the root. That file was used before the node template was restructured into `node`, `runtime`, and `pallets` directories. The new home for this file is `node/build.rs` so this one can be safely deleted.


JoshOrndorff:
More info. This file was already removed from the upstream node template. We just missed it here because the process of updating was to copy new files over old ones.

PR #31, opened 2020-04-07
Update Node Template CLI  to use new SubstrateCLI

4meta5:
There are still two errors. The first one is one that I encountered when updating my node so I tried updating this node and am getting both of these errors now.

```rust
error: environment variable `SUBSTRATE_CLI_IMPL_VERSION` not defined
  --> node/src/command.rs:29:3
   |
29 |         env!("SUBSTRATE_CLI_IMPL_VERSION")
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0432]: unresolved import `sc_cli::SubstrateCli`
  --> node/src/command.rs:20:5
   |
20 | use sc_cli::SubstrateCli;
   |     ^^^^^^^^^^^^^^^^^^^^ no `SubstrateCli` in the root

error: aborting due to 2 previous errors
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,9 +99,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.27"
+version = "1.0.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "013a6e0a2cbe3d20f9c60b65458f7a7f7a5e636c5d0f45a5a6aee5d4b1f01785"
+checksum = "d9a60d744a80c30fcb657dfe2c1b22bcb3e814c1a1e3674f32bf5820b570fbff"
 
 [[package]]
 name = "app_dirs"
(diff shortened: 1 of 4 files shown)


JoshOrndorff:
It looks like you haven't updated the dependencies in `Cargo.toml`. The new CLI isn't included in `v2.0.0-alpha.5` that the template depends on.

4meta5:
I think I really understood and still don't get how the versioning works. This file has `2.0.0-alpha.5` https://github.com/paritytech/substrate/blob/master/bin/node/cli/Cargo.toml and this is the crate it depends on?

Also, another dumb question, why don't the dependencies require `git = 'https://github.com/paritytech/substrate.git'`. Where is that implied?

JoshOrndorff:
> I still don't get how the versioning works. This file has 2.0.0-alpha.5 https://github.com/paritytech/substrate/blob/master/bin/node/cli/Cargo.toml and this is the crate it depends on?

The packages in the paritytech/substrate repo still say alpha.5 and have said that ever since alpha.5 was released. The process they've been following is to update the package version in the `Cargo.toml` files right before they cut a new Substrate release. So whenever they are ready to release alpha.6, they will update the `Cargo.toml` files to say alpha.6, then tag the release.

> why don't the dependencies require `git = 'https://github.com/paritytech/substrate.git'`. Where is that implied?

That's a great question, and it recently changed. This node-template repo actually _doesn't_ depend on paritytech/substrate github repository anymore. Rather it pulls its dependencies from crates.io. So when you see a dependency like [this one](https://github.com/substrate-developer-hub/substrate-node-template/blob/4be00bdec6fd3aef6448d4613559522a84e22866/node/Cargo.toml#L41-L42)
``toml
[dependencies.sc-client]
version = '0.8.0-alpha.5'
```
We're actually depending on https://crates.io/crates/sc-client.

The new CLI is not yet in a release and is not yet published on crates.io. So if you want to use it right now you'll have to switch the dependencies back to something like.
```toml
[dependencies.sc-client]
git = 'https://github.com/paritytech/substrate.git'
rev = '11493abe7d052c7465fa7760c57824fb455e0e68' # Or whatever
```

4meta5:
That makes sense. Thank you for the comprehensive explanation!

PR #34, opened 2020-05-01
Add note about versions to README

danforbes:
This is a kind of template for a change that I believe we should make to a number of related repositories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,13 +18,8 @@ undergo many breaking changes, the Substrate team takes great pains to release c
 of related codebases as close to simultaneously as possible. This helps ensure a positive
 experience for Substrate developers and also reduces the support overhead on the Substrate team.
 Please always be sure that you are using specific, compatible versions of the components in your
-Substrate development environment. The versions this repository that are currently available are:
-
-- [v2.0.0-alpha.6](https://github.com/substrate-developer-hub/substrate-node-template/tree/v2.0.0-alpha.6)
-- [v2.0.0-alpha.5](https://github.com/substrate-developer-hub/substrate-node-template/tree/v2.0.0-alpha.5)
-- [v2.0.0-alpha.3](https://github.com/substrate-developer-hub/substrate-node-template/tree/v2.0.0-alpha.3)
-- [pre-v2.0-3e65111](https://github.com/substrate-developer-hub/substrate-node-template/tree/pre-v2.0-3e65111)
-- [v1.5.0](https://github.com/substrate-developer-hub/substrate-node-template/tree/v1.5.0)
+Substrate development environment. You can find the list of all the available versions at:
+https://github.com/substrate-developer-hub/substrate-node-template/tags
 
 ## Build
 
--- a/README.md
+++ b/README.md
@@ -7,6 +7,25 @@ A new Substrate node, ready for hacking. This node includes:
 * Aura block authoring
 * Grandpa finality gadget
 
+## Version
+
+You are on the `master` branch of this repository! In general, the only reason to checkout and work
+with this branch is if you are making changes to the Substrate Node Template itself. If you are
+using the code in this repository to follow a tutorial or some other type of guide, you should
+checkout one of the tagged versions of this repository that corresponds with a tagged version of
+the Substrate codebase itself. Because Substrate is such a fast-moving project that is likely to
+undergo many breaking changes, the Substrate team takes great pains to release compatible versions
+of related codebases as close to simultaneously as possible. This helps ensure a positive
+experience for Substrate developers and also reduces the support overhead on the Substrate team.
+Please always be sure that you are using specific, compatible versions of the components in your
+Substrate development environment. The versions this repository that are currently available are:
+
+- [v2.0.0-alpha.6](https://github.com/substrate-developer-hub/substrate-node-template/tree/v2.0.0-alpha.6)
+- [v2.0.0-alpha.5](https://github.com/substrate-developer-hub/substrate-node-template/tree/v2.0.0-alpha.5)
+- [v2.0.0-alpha.3](https://github.com/substrate-developer-hub/substrate-node-template/tree/v2.0.0-alpha.3)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,6 +7,25 @@ A new Substrate node, ready for hacking. This node includes:
 * Aura block authoring
 * Grandpa finality gadget
 
+## Version
+
+You are on the `master` branch of this repository! In general, the only reason to checkout and work
+with this branch is if you are making changes to the Substrate Node Template itself. If you are
+using the code in this repository to follow a tutorial or some other type of guide, you should
+checkout one of the tagged versions of this repository that corresponds with a tagged version of
```
joepetrowski, line 15:
```suggestion
You are on the `master` branch of this repository! In general, the only reason to check out and work
with this branch is if you are making changes to the Substrate Node Template itself. If you are
using the code in this repository to follow a tutorial or some other type of guide, you should
check out one of the tagged versions of this repository that corresponds with a tagged version of
```

danforbes: Wasn't sure about that one thanks much :pray:

README.md
```diff
@@ -7,6 +7,25 @@ A new Substrate node, ready for hacking. This node includes:
...
+of related codebases as close to simultaneously as possible. This helps ensure a positive
+experience for Substrate developers and also reduces the support overhead on the Substrate team.
+Please always be sure that you are using specific, compatible versions of the components in your
+Substrate development environment. The versions this repository that are currently available are:
+
+- [v2.0.0-alpha.6](https://github.com/substrate-developer-hub/substrate-node-template/tree/v2.0.0-alpha.6)
+- [v2.0.0-alpha.5](https://github.com/substrate-developer-hub/substrate-node-template/tree/v2.0.0-alpha.5)
+- [v2.0.0-alpha.3](https://github.com/substrate-developer-hub/substrate-node-template/tree/v2.0.0-alpha.3)
+- [pre-v2.0-3e65111](https://github.com/substrate-developer-hub/substrate-node-template/tree/pre-v2.0-3e65111)
+- [v1.5.0](https://github.com/substrate-developer-hub/substrate-node-template/tree/v1.5.0)
```
riusricardo, line 27:
```suggestion
Substrate development environment. You can find the list of all the available versions at:
https://github.com/substrate-developer-hub/substrate-node-template/tags
```

riusricardo: I would keep it more general to remove the need of updating it after tagging a new version.

danforbes: You da real MVP :pray:

JoshOrndorff:
I think warning about the versioning scheme is a good idea, and I like the language you've used. I wonder whether it makes sense to put _in_ the version control system like this though.

What is the procedure for when alpha.7 comes out? Of course we will copy the new code into this repo as always, but if we then when we tag `v2.0.0-alpha.7` it will contain the text "you are on the master branch".

The best process I can think of would be:
* copy in the new code
* revert the commit that add the warning
* tag the version
* revert the revert thus restoring the warning

danforbes:
@JoshOrndorff, great question - admittedly I had not thought that all the way through! Here's my proposal for a branching strategy that I believe supports a clearer focus on versioning:

- `master` branch is our development branch. Tagged versions are not on `master`.
- New versions start by creating a branch off `master`.
- Changes are made to the version branch to upgrade to the new version and this is pulled into `master`.
- An additional commit is made on the version branch where we can do things like updating the README, etc - this commit is what is tagged and eventually released as the version. At this point, we can delete the version branch and just leave the dangling tag.

This is just one suggested approach. No rush to get this merged in now...just spent some time thinking about this as I was preparing for the `alpha.8` release.

JoshOrndorff:
What about just changing the text slightly and removing the phrase, "You are on the  master branch". Most of what you wrote there is of the "heres how you _should_ use this repo" variety which I think is appropriate on any branch. That would save us from having to remove that phrase during each release, and would allow tagged version to still be in master branch.

Might also be worth mentioning in the text that a common use of the template is cloning it to start your own chain. The currently proposed text only mentions following guides.

JoshOrndorff:
I'm also fine proceeding with the plan you described if you prefer it @danforbes :+1:

danforbes:
Is there a compelling reason to have tagged versions on `master` or is it just convention? Is it even convention, or is it just something that some people feel comfortable with? Honest questions that I don't know the answers to.

JoshOrndorff:
Having a shared history between tagged versions allows anyone to easily answer the question, "What are the differences between these two versions".

This repo has traditionally not had an "active development branch". Active development has happened in paritytech/substrate/master, and in https://github.com/substrate-developer-hub/substrate-node-template/pull/44#issuecomment-629667937 you suggested you would like that to continue.

danforbes:
You can compare across tags that are not on the same branch; there is still a shared history between them https://github.com/substrate-developer-hub/substrate-pallet-template/compare/v2.0.0-alpha.7...v2.0.0-alpha.8 What's interesting to me is that the diff in the link doesn't look like I'd expect it to; it seems to show an upgrading happening from alpha.6 to alpha.8 and to be honest, I'm not entirely clear why. If the strategy I'm suggesting makes it difficult to get meaningful diffs, I agree that it is a compelling reason not to use it. I'm inclined to believe there's a good answer to this question and a good strategy for addressing it.

JoshOrndorff:
This is the first I'm noticing that the alpha.7 and alpha.8 tags are _not_ in the history of master. The diff appears to show an update from alpha.6 because alpha.6 is the common ancestor of the two tags you're comparing.

I'm guessing that when those alpha.7 and alpha.8 tags were created the process was:
1. checkout master
2. update code
3. open PR
4. tag commit in pr branch
5. squash-merge PR into master

When I tagged alpha.6 my process was:

1. Checkout master
2. update code
3. open PR
4. squash merge PR into master
5. tag commit in master.

danforbes:
Yes, you're absolutely correct and this is probably a reflection of some kind of naivete with respect to git on my part.

danforbes:
This is really stale. We can revisit this conversation later if it's something that becomes important in the future.

PR #37, opened 2020-05-06
alpha.6 to github

riusricardo:
If merged, it requires to push the tag to this commit.

Fixes incorrect alpha.7 dependencies.

### DIFF
--- a/pallets/template/Cargo.toml
+++ b/pallets/template/Cargo.toml
@@ -18,23 +18,29 @@ package = 'parity-scale-codec'
 version = '1.3.0'
 
 [dependencies.frame-support]
+git = 'https://github.com/paritytech/substrate.git'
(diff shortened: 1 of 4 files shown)


pallets/template/Cargo.toml
```diff
@@ -18,23 +18,29 @@ package = 'parity-scale-codec'
 version = '1.3.0'
 
 [dependencies.frame-support]
+git = 'https://github.com/paritytech/substrate.git'
 default-features = false
-version = '2.0.0-alpha.6'
+tag ='v2.0.0-alpha.6'
```
danforbes, line 23: Formatting: should there be a space between `=` and the tag name?

riusricardo: right

PR #41, opened 2020-05-13
Use docker to setup a local dev chain 

kaichaosun:
I recently need to help solve the variance of platforms when developing on Substrate.
Docker is still the first choice when dealing with this situation. 

It would be good to know how are we going to support different platforms in the long term.

This change works in my MacOS, but haven't tested it in other platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ If you just want to run the compiled binary,
 ./scripts/docker_run.sh ./target/release/node-template --dev --ws-external
 ```
 
+Other commands are similar. Let's try purge the local dev chain here:
+
+```bash
+./scripts/docker_run.sh ./target/release/node-template purge-chain --dev
+```
+
 You can also check whether the code is able to compile or not,
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Additional CLI usage options are available and may be shown by running `cargo ru
 
 ### Run in Docker
 
-Install Docker first, then run following command to start a single node development chain.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -80,6 +80,14 @@ cargo run -- \
 
 Additional CLI usage options are available and may be shown by running `cargo run -- --help`.
 
+### Run in Docker
+
+Install [Docker](https://docs.docker.com/get-docker/) first, then run following command to start a single node development chain.
```
joepetrowski, line 85:
```suggestion
Install [Docker](https://docs.docker.com/get-docker/) first, then run the following command to start a single node development chain.
```

jimmychu0807:
I tried on ubuntu v16 and it runs also. A few comments:

1. On the last command in `docker-compose.yml` we are combining `cargo build` and running the node-template together. Can we have a shortcut cmd to just run the node-template? Because cargo build with recompiling often triggered (I think it checks the file timestamp instead of the file content for cache), this often cause unnecessary recompiling and I waiting for 5-6 mins.

2. Can we also provide a shortcut to do `purge-chain --dev` cmd?

kaichaosun:
@jimmychu0807 
I add more commands to the readme, like purge chain, cargo check, only run the compiled node.
Source file changes will trigger rebuild, but other than that it's using the cache during my test.

jimmychu0807:
I see. When you pass your own command to `docker-compose run`, it replaces the command specified in `docker-compose.yml`.

Also I realize since the storage in docker is ephemeral, we don't need to purge the chain. The chain info is just gone when the container is tear down. You also remove it when running `docker_run.sh`.

PR #44, opened 2020-05-15
Upgrade to Substrate v2.0.0-alpha.8

riusricardo:
PR to merge @danforbes's update to alpha.8

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,6 @@
 # Substrate Node Template
 
-A new Substrate node, ready for hacking. This node includes:
-
-* A FRAME-based runtime
-* A template pallet
-* Aura block authoring
-* Grandpa finality gadget
+A new FRAME-based Substrate node, ready for hacking.
 
 ## Build
 
@@ -61,7 +56,7 @@ cargo run -- \
   --chain=local \
   --alice \
   --node-key 0000000000000000000000000000000000000000000000000000000000000001 \
-  --telemetry-url ws://telemetry.polkadot.io:1024 \
+  --telemetry-url 'ws://telemetry.polkadot.io:1024 0' \
   --validator
 ```
 
@@ -74,7 +69,7 @@ cargo run -- \
   --chain=local \
   --bob \
   --port 30334 \
-  --telemetry-url ws://telemetry.polkadot.io:1024 \
+  --telemetry-url 'ws://telemetry.polkadot.io:1024 0' \
   --validator
 ```
 
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
-authors = ['Anonymous']
+authors = ['Substrate DevHub <https://github.com/substrate-developer-hub>']
 build = 'build.rs'
-description = 'Substrate Node template'
+description = 'Substrate node template'
 edition = '2018'
-homepage = 'https://substrate.dev'
+homepage = 'https://substrate.io'
 license = 'Unlicense'
 name = 'node-template'
-repository = 'https://github.com/paritytech/substrate/'
-version = '2.0.0-alpha.7'
+repository = 'https://github.com/substrate-developer-hub/substrate-node-template/'
+version = '2.0.0-alpha.8'
 
 [package.metadata.docs.rs]
 targets = ['x86_64-unknown-linux-gnu']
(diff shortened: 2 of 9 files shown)


riusricardo:
@danforbes The code is correct. You just need to push the final lock file.

node/Cargo.toml
```diff
@@ -1,107 +1,116 @@
...
-description = 'Substrate Node template'
+description = 'Substrate node template'
 edition = '2018'
-homepage = 'https://substrate.dev'
+homepage = 'https://substrate.io'
 license = 'Unlicense'
 name = 'node-template'
-repository = 'https://github.com/paritytech/substrate/'
-version = '2.0.0-alpha.7'
+repository = 'https://github.com/substrate-developer-hub/substrate-node-template/'
```
riusricardo, line 9: Same as above. If we change it it should match upstream. The reason for this is that is reduces the maintenance and the time it takes to update the next version.

node/Cargo.toml
```diff
@@ -1,107 +1,116 @@
 [package]
-authors = ['Anonymous']
+authors = ['Substrate DevHub <https://github.com/substrate-developer-hub>']
```
riusricardo, line 2: If we plan to keep this change it should go upstream. IMHO I don't think that we should keep it.
It is a change that we may not want to have on the main Substrate repo.

node/Cargo.toml
danforbes: Personally I think that the value derived from updating this line is worth the negligible amount of time it takes to do so. I think it will help reduce confusion among Substrate developers. In particular, I think we want to make it clear that there is a distinction between Substrate and the node template.

node/Cargo.toml
danforbes: Maybe we should change it to match the other components of the Substrate repository, then: https://github.com/paritytech/substrate/blob/master/frame/assets/Cargo.toml#L4

danforbes: So I vote on reverting this change from this PR and making an upstream change to bring the node template in line with the other crates in the Substrate repository; it should list Parity Technologies as the author.

danforbes: I started taking a look at the node template that ships with Substrate core and I no longer believe we should make this change upstream. I think it's good for us to think about and promote this repo as if it's a separate project that our team maintains, so I think making these minor adjustments to the output of the node template script is worth it.

README.md
```diff
@@ -1,11 +1,6 @@
 # Substrate Node Template
 
-A new Substrate node, ready for hacking. This node includes:
-
-* A FRAME-based runtime
-* A template pallet
-* Aura block authoring
-* Grandpa finality gadget
+A new FRAME-based Substrate node, ready for hacking.
```
JoshOrndorff, line 3: I like the detail that got removed here. I also don't think it makes a ton of sense to describe the node as "FRAME-based". I think that would make more sense for the runtime. Not a blocker, just my two-cents.

node/Cargo.toml
JoshOrndorff: I agree with both of these statements:
* > If we change it it should match upstream
* >  I think we want to make it clear that there is a distinction between Substrate and the node template.

Maybe it's time that DevHub make the decision that the node template is ready to graduate from the mono-repo to living on its own here. Of course then we'll be responsible for maintaining it.

node/Cargo.toml
JoshOrndorff: I think the reason it was previously anonymous is because it is a _template_. Whoever clones it should put their own name in there, and in the case they forget to do that, we might not want our name on their work.

I really don't care if we stick to that logic though.

node/Cargo.toml
JoshOrndorff: This conversation shouldn't block this PR though. How about this plan:

We include Dan's change in the alpha.8 release, and meanwhile we decide the fate of the upstream node template.:
* It get's killed. The node template only lives here in devhub
* It lives on and also has the url for the devhub repo
* It lives on keeping the main substrate repo

node/Cargo.toml
JoshOrndorff: I'm fine with any of the options suggested above.

danforbes:
My proposal is that we continue to use the node template provided by the team of core Substrate developers and that we make _very_ minor modifications to it in order to create the Substrate DevHub Node Template. I think that this pattern reinforces the mental model of the DevHub as the first-line of Substrate power-users.

JoshOrndorff:
If we decide to keep the upstream node template, I think we should have a clear and common understanding of what changes are made upstream and what changes are made here. IMO the primary value of having an upstream node template in the main Substrate repo is that when core devs make breaking changes, they are forced to update the node template accordingly. If our node template diverges from upstream we lose that benefit.

danforbes:
Agree completely - we should not change the logic of the code _at all_. We should only change things like the author, repository, etc.

PR #62, opened 2020-06-26
Upgrade to v2.0.0-rc4

danforbes:
Please note the changes I made to the `README`. cc @jeluard :cartwheeling:

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ curl https://getsubstrate.io -sSf | bash -s -- --fast
 
 ### Manual Method
 
-Simple steps for Linux-based systems can be found below; you can
+Manual steps for Linux-based systems can be found below; you can
 [find more information at substrate.dev](https://substrate.dev/docs/en/knowledgebase/getting-started/#manual-installation).
 
 Install Rust:
--- a/README.md
+++ b/README.md
@@ -2,9 +2,25 @@
 
 # Substrate Node Template
 
-A new FRAME-based Substrate node, ready for hacking.
+A new FRAME-based Substrate node, ready for hacking :rocket:
 
-## Build
+## Local Development
+
+Follow these steps to prepare your local environment for Substrate development :hammer_and_wrench:
+
+### Simple Method
+
+You can install all the required dependencies with a single command (be patient, this can take up
+to 30 minutes).
+
+```bash
+curl https://getsubstrate.io -sSf | bash -s -- --fast
+```
+
+### Manual Method
+
+Simple steps for Linux-based systems can be found below; you can
+[find more information at substrate.dev](https://substrate.dev/docs/en/knowledgebase/getting-started/#manual-installation).
 
 Install Rust:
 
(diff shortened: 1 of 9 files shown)


jeluard:
Best 👏 changes 👏 ever 👏

README.md
```diff
@@ -2,9 +2,25 @@
...
+You can install all the required dependencies with a single command (be patient, this can take up
+to 30 minutes).
+
+```bash
+curl https://getsubstrate.io -sSf | bash -s -- --fast
+```
+
+### Manual Method
+
+Simple steps for Linux-based systems can be found below; you can
```
JoshOrndorff, line 22:
```suggestion
Manual steps for Linux-based systems can be found below; you can
```

README.md
```diff
@@ -40,54 +66,28 @@ Start a development chain with:
 ./target/release/node-template --dev
 ```
 
-Detailed logs may be shown by running the node with the following environment variables set: `RUST_LOG=debug RUST_BACKTRACE=1 cargo run -- --dev`.
+Detailed logs may be shown by running the node with the following environment variables set:
+`RUST_LOG=debug RUST_BACKTRACE=1 cargo run -- --dev`.
```
JoshOrndorff, line 70: I'm fine with this, but it should be changed upstream as well because this content comes from there.

README.md
```diff
@@ -40,54 +66,28 @@ Start a development chain with:
...
-  --chain=local \
-  --bob \
-  --port 30334 \
-  --telemetry-url 'ws://telemetry.polkadot.io:1024 0' \
-  --validator
-```
-
-Additional CLI usage options are available and may be shown by running `cargo run -- --help`.
+If you want to see the multi-node consensus algorithm in action, refer to
+[our Start a Private Network tutorial](https://substrate.dev/docs/en/tutorials/start-a-private-network/).
```
JoshOrndorff, line 75: Same, why not change this all upstream?

README.md
```diff
@@ -99,29 +99,3 @@ This command will firstly compile your code, and then start a local development
...
-git checkout <branch/tag/sha1>
-
-# Run the helper script to generate a node template.
-# This script compiles Substrate and takes a while to complete. It takes a relative file path
-#   from the current dir. to output the compressed node template.
-.maintain/node-template-release.sh ../node-template.tar.gz
-```
-
-Noted though you will likely get faster and more thorough support if you stick with the releases
-provided in this repository.
```
JoshOrndorff, line 127: This is gonna become very important if we want people who are using the node template as a starting point to be able to use cumulus.

README.md
danforbes: Yes, this is a really good point and something I've been meaning to address. However, I think we may want to consider the documentation for the Substrate DevHub Node Template to be one of the places it diverges the most from the upstream node template that ships with the core Substrate code base.

README.md
danforbes: I think this is one place where we want to refer back to our own content in the version of this template that we maintain. I can understand why the maintainers of the core Substrate code base would prefer to provide this information inline in a more terse form.

README.md
danforbes: Good point, so we should consider including this documentation with the Cumulus code base. When it comes to the node template, though, I'm not too keen on documenting capabilities and then saying we don't really support them. This documentation is something that we may also want to consider contributing back upstream, but I would still think we want to remove it from the README for the Substrate DevHub Node Template.

README.md
wheresaddie: yeap totally agree with you @danforbes

README.md
wheresaddie: I would suggest we remove it from the README for the Substrate DevHub Node Template as it's something that would decrease our maintenance burden as people migrate to cumulus

PR #66, opened 2020-07-27
Remove unused dependencies

c410-f3r:


danforbes:
Thank you so much for this extremely helpful PR, @c410-f3r :pray: Looking forward to seeing what you cook up next :yum:

Issue #67, opened 2020-07-28
 Remove References to Out-of-Date get-substrate.sh Script


danforbes:
Ref: paritytech/scripts#218

PR #69, opened 2020-08-06
Add set-up instructions

danforbes:
Relates to #67

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -1,0 +1,48 @@
+# Substrate Developer Set-Up Guide
+
+The purpose of this document is to provide developers with the information they need to set-up a Substrate development
+environment. For each operating system described below, a standard Docker container has been used to verify that the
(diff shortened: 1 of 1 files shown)


JoshOrndorff:
I'm not sure if this is done, but I think MacOS is a supported platform, right?

And when this gets merged we should probably remove https://substrate.dev/docs/en/knowledgebase/getting-started/ and https://substrate.dev/docs/en/knowledgebase/getting-started/windows-users

wheresaddie (approved):

danforbes:
This PR should actually probably be closed because I think we decided this material would live in the Knowledge Base https://substrate.dev/docs/en/knowledgebase/getting-started/

wheresaddie:
yeap sorry about that, I realized it after I pushed the changes it was dead already :)

PR #76, opened 2020-08-23
Update setup.md

wheresaddie:
minor edits

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -1,20 +1,19 @@
 # Substrate Developer Set-Up Guide
 
-The purpose of this document is to provide developers with the information they need to set-up a Substrate development
-environment. For each operating system described below, a standard Docker container has been used to verify that the
-provided steps are sufficient to build and run the template node in this repository.
+The purpose of this document is to provide developers of all levels the information they need to set-up a Substrate development
+environment. For each operating system, a standard Docker container has been used to verify that the
+provided steps successfully build and run the node template in this repository. The estimated set up time is 45 minutes.
 
 ## Ubuntu
 
-This set-up was verified using [the standard Ubuntu Docker container](https://hub.docker.com/_/ubuntu). This container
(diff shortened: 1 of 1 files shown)


setup.md
```diff
@@ -1,20 +1,19 @@
 # Substrate Developer Set-Up Guide
 
-The purpose of this document is to provide developers with the information they need to set-up a Substrate development
-environment. For each operating system described below, a standard Docker container has been used to verify that the
-provided steps are sufficient to build and run the template node in this repository.
+The purpose of this document is to provide developers of all levels the information they need to set-up a Substrate development
+environment. For each operating system, a standard Docker container has been used to verify that the
+provided steps successfully build and run the node template in this repository. The estimated set up time is 45 minutes.
```
JoshOrndorff, line 5: "in this repository" is left over from when Dan drafted this in the node template repo. It should probably be a link to the node template at this point.

Also that setup time seems long (at least on linux). I previously recorded a video of that script executing in <3min/ Although I know it's longer on osx.

setup.md
```diff
@@ -44,5 +43,4 @@ rustup update
 
 ## Unsupported Operating Systems
 
-Any operating system not explicitly described by this document should be considered unsupported for Substrate
-development.
+Any operating system not explicitly described by this document should be considered unsupported by Substrate.
```
JoshOrndorff, line 46: I think this should be more encouraging. Something like.

Substrate can and does run on a wider variety of platforms than those listed here. However, platforms not listed above are not considered officially supported by the Substrate Developer Hub.

setup.md
wheresaddie: good point @JoshOrndorff I tested it on win/osx machines and it was closer to 30-45 mins.. on my ubuntu it was just under 5. . so perhaps this can be removed.

PR #81, opened 2020-09-08
Fix bug where the wrong import queue was used in service

### LINKED ISSUE #80: Self-authored blocks use partial import queue
Because of an upstream bug that exists when the rc6 node template was cut, when a node authors a block, it only passes that block through part of the import queue (the grandpa part, but not the aura part). That bug was fixed upstream https://github.com/paritytech/substrate/pull/7014

Typically new node templates are cut only when new upstream versions are tagged, but in this case, because we have a user-facing bug in the codebase and it will be copied into each new template fork, I recommend we cut a new release right away. Perhaps it could be called `v2.0.0-rc6+1`.

danforbes:
Backport paritytech/substrate#7014
Closes #80

### DIFF
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -27,7 +27,12 @@ pub fn new_partial(config: &Configuration) -> Result<sc_service::PartialComponen
 	sp_consensus::DefaultImportQueue<Block, FullClient>,
 	sc_transaction_pool::FullPool<Block, FullClient>,
 	(
-		sc_finality_grandpa::GrandpaBlockImport<FullBackend, Block, FullClient, FullSelectChain>,
+		sc_consensus_aura::AuraBlockImport<
+			Block,
+			FullClient,
+			sc_finality_grandpa::GrandpaBlockImport<FullBackend, Block, FullClient, FullSelectChain>,
+			AuraPair
+		>,
 		sc_finality_grandpa::LinkHalf<Block, FullClient, FullSelectChain>
 	)
 >, ServiceError> {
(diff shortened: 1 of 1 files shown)


JoshOrndorff (approved):
Looks good. But there are some process questions:

* How are we going to tag this?
* Are we going to update tutorials?

node/src/service.rs
```diff
@@ -27,7 +27,12 @@ pub fn new_partial(config: &Configuration) -> Result<sc_service::PartialComponen
 	sp_consensus::DefaultImportQueue<Block, FullClient>,
 	sc_transaction_pool::FullPool<Block, FullClient>,
 	(
-		sc_finality_grandpa::GrandpaBlockImport<FullBackend, Block, FullClient, FullSelectChain>,
+		sc_consensus_aura::AuraBlockImport<
+			Block,
+			FullClient,
+			sc_finality_grandpa::GrandpaBlockImport<FullBackend, Block, FullClient, FullSelectChain>,
+			AuraPair
+		>,
```
JoshOrndorff, line 35: This is a great opportunity to learn about the onion-like nature of the block import pipeline, which is described more fully at https://substrate.dev/docs/en/knowledgebase/advanced/block-import .

The complete block import pipeline is `Grandpa -> Aura -> Client`. Before this PR fixed it, only the `Aura -> Client` portion was used. The snippet I've highlighted here removes the type of the partial pipeline and replace it with the type of the complete pipeline. Notice that the removed line 30 is identical to the added line 33. that shows the partial pipeline nested inside the complete pipiline. Also notice that the line 33 includes the type `FullClient`, showing the final layer of wrapping.

riusricardo (approved):
Nice! 
>How are we going to tag this?

IMO, we can push the tag. The dependencies are still pointing to rc6 tag. I don't know if it will break any previous versions.

Issue #90, opened 2020-10-08
first time install failed, error[E0282]: type annotations needed


Maar-io:
**Description**

I did manage to install the node and run it together with the web front-end maybe 5 days ago. But when I started all over again I can't get it compiled. I did to reinstall rust and de-brew dependency modules but all in vain. Also tried with rustc 1.46, restarted mac, reinstalled cmake, openssl...
Attached is the log with the whole installation procedure.
[subrtrate-install.log](https://github.com/substrate-developer-hub/substrate-node-template/files/5351338/subrtrate-install.log)

**Steps to Reproduce**

following step by step [Create Your First Substrate Chain](https://substrate.dev/docs/en/tutorials/create-your-first-substrate-chain/setup)

**Environment**
- Operating system: MacOS, Kernel Version: Darwin 18.2.0
- Template version/tag: v2.0.0
- Rust version (run `rustup show`):
```
Default host: x86_64-apple-darwin
rustup home:  /Users/mario/.rustup

installed toolchains

stable-x86_64-apple-darwin (default)
nightly-x86_64-apple-darwin

active toolchain

stable-x86_64-apple-darwin (default)
rustc 1.47.0 (18bf6b4f0 2020-10-07)
```
**Logs, Errors or Screenshots**
```
...
   Compiling frame-support-procedural v2.0.0
   Compiling libp2p-core v0.22.1
error: failed to run custom build command for `node-template-runtime v2.0.0 (/Users/mario/projects/substrate-node-template/runtime)`

Caused by:
  process didn't exit successfully: `/Users/mario/projects/substrate-node-template/target/release/build/node-template-runtime-cae9ad6029c9f681/build-script-build` (exit code: 1)
  --- stdout
  Executing build command: "rustup" "run" "nightly" "cargo" "rustc" "--target=wasm32-unknown-unknown" "--manifest-path=/Users/mario/projects/substrate-node-template/target/release/wbuild/node-template-runtime/Cargo.toml" "--color=always" "--release"

  --- stderr
      Updating crates.io index
   Downloading crates ...
    Downloaded getrandom v0.1.15
    Downloaded substrate-wasm-builder v2.0.0
```

**This error comes after random file and doesn't look connected with libp2p-core v0.22.1 in this case.
After continuing with compile it always breaks with following error:**

```
...     
     Compiling sp-wasm-interface v2.0.0
     Compiling sp-arithmetic v2.0.0
     Compiling finality-grandpa v0.12.3
[...]

PR #91, opened 2020-10-09
Update setup.md

wheresaddie:
update setup for ubtunu to fix rust nightly bug which breaks the node template setup #90


danforbes:
I was thinking that we should do something like this, but I think we decided we were not going to include these set up instructions in the node template.

wheresaddie:
yeap totally agree with you @danforbes just was trying to find a patch to address the sudden slew of errors people were getting in the meantime

PR #99, opened 2020-10-20
Remove rust-toolchain file

athei:
Architecture specific toolchains are [not allowed](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) in this file. We can assume that stable is the default toolchain and should not dictate with which toolchain the user builds the client. Also this breaks under macOS.

We are only interested with which toolchain the runtime is build. For this we should set the `WASM_BUILD_TOOLCHAIN` variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Once the development environment is set up, build the node template. This comman
 [native](https://substrate.dev/docs/en/knowledgebase/advanced/executor#native-execution) code:
 
 ```bash
-cargo build --release
+WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo build --release
 ```
 
 ## Run
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ services:
       - "9944:9944"
     environment:
       - CARGO_HOME=/var/www/node-template/.cargo
+      - WASM_BUILD_TOOLCHAIN=nightly-2020-10-05
     volumes:
       - .:/var/www/node-template
       - type: bind
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,2 +1,0 @@
-stable-x86_64-unknown-linux-gnu
-nightly-2020-10-05


riusricardo (approved):
Thanks @athei, I also had my doubts about the rust-toolchain file. Thanks for the link :+1:

athei:
Yeah the wording needs to change, too:

> Developers that are building with Substrate (as opposed to the developers building Substrate itself) should use a specific Rust nightly version that is known to be compatible with the version of Substrate they are using. For instance, The Substrate Developer Hub Node Template includes a toolchain file that specifies the Rust nightly version to use. Use Rustup to install the correct nightly:

Why the distinction between people building substrate and building with substrate? I guess the build can also fail when you are using substrate merely as a dependency. I guess it depends on which pallets you are using.

danforbes:
@athei can you suggest an update to the wording? The distinction you're inquiring about is clarified further in the docs:
> Developers that are building Substrate _itself_ should always uses the latest version of Rust stable and nightly for compilation.

Developers that are working _with_ Substrate should use a supported version of Substrate that is known to work with a specific nightly. Developers that are working _on_ Substrate should always use latest nightly. That is my understanding at this time.

athei:
I am working _on_ substrate and I of course I cannot use the latest nightly because it simply does not build because of a regression introduced in nightly 2020-10-06. We are all in the same boat here. I think for an introductory document it is important to pin it to a working nightly version. Nothing is more frustrating than trying out a new framework you discovered just to find out that it doesn't even build.

I find this distinction confusing. I recommend just leaving this notion out and write the tutorial so that it uses a pinned nightly version which we can update from time to time. 

Then add a text box that explains **why** we are pinning (because new nightlies are allowed to break existing code) and that the user can try the latest nightly but no guarantees.

wheresaddie:
> Then add a text box that explains **why** we are pinning (because new nightlies are allowed to break existing code) and that the user can try the latest nightly but no guarantees.

hi @athei - thanks for all this input. If you want to modify text and add this additional information/box/pinds, and cc us in the PR, more than willing to make adjustments assuming they don't break existing docs. thanks!

riusricardo:
I'm gonna merge this to fix the Docker and hopefully the playground build

athei:
> > Then add a text box that explains **why** we are pinning (because new nightlies are allowed to break existing code) and that the user can try the latest nightly but no guarantees.
> 
> hi @athei - thanks for all this input. If you want to modify text and add this additional information/box/pins, and cc us in the PR, more than willing to make adjustments assuming they don't break existing docs. thanks!

I will post my proposal as PR soonish.

PR #100, opened 2020-10-23
Nix shell sync with Substrate

kaichaosun:
Also fixed error when using macos.
https://github.com/paritytech/substrate/pull/7376

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -15,6 +15,8 @@ with nixpkgs; pkgs.mkShell {
     cmake
     pkg-config
     rust-nightly
+  ] ++ stdenv.lib.optionals stdenv.isDarwin [
+    darwin.apple_sdk.frameworks.Security
   ];
 
   LIBCLANG_PATH = "${llvmPackages.libclang}/lib";
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,23 @@
+let
+  mozillaOverlay =
+    import (builtins.fetchGit {
+      url = "https://github.com/mozilla/nixpkgs-mozilla.git";
+      rev = "57c8084c7ef41366993909c20491e359bbb90f54";
+    });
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -0,0 +1,25 @@
+let
+  mozillaOverlay =
+    import (builtins.fetchGit {
+      url = "https://github.com/mozilla/nixpkgs-mozilla.git";
+      rev = "57c8084c7ef41366993909c20491e359bbb90f54";
+    });
+  nixpkgs = import <nixpkgs> { overlays = [ mozillaOverlay ]; };
+  rust-nightly = with nixpkgs; ((rustChannelOf { date = "2020-10-01"; channel = "nightly"; }).rust.override {
```
riusricardo, line 8: Is it possible to use the same version as docker, 2020-10-05?

riusricardo (commented):
Will it work if we move it to the scripts folder?

riusricardo:
Ok, I just saw your other PR. It needs to be in the root folder.

shell.nix
pepyakin: meanwhile it was already upgraded : )

shell.nix
```diff
@@ -0,0 +1,25 @@
...
+in
+with nixpkgs; pkgs.mkShell {
+  buildInputs = [
+    clang
+    cmake
+    pkg-config
+    rust-nightly
+  ] ++ stdenv.lib.optionals stdenv.isDarwin [
+    darwin.apple_sdk.frameworks.Security
+  ];
```
pepyakin, line 20: care to upstream this?

shell.nix
kaichaosun: node-template still depends on v2.0.0, i believe latest working nightly is 2020-10-05.

PR #105, opened 2020-11-07
Update Local Development docs to use Makefile

danforbes:


athei:
A makefile that only contains phony targets is nothing but a glorified shell script. If you **must** use a Makefile I would at least flag those targets as phony.

That said, I strongly recommend **not** using a makefile for a rust project. I usually get repulsed by projects that do because that usually means they did not manage to get their project to build with cargo alone.

Maybe you can accomplish the same thing with cargo aliases or cargo-xtask. Those will scare Rust developers to a lesser degree than makefiles I think.

kaichaosun:
Nix shell is known for its reliable and reproducible environment across different machines.  why delete it?

danforbes:
@athei I was not able to able to figure out a way to use Cargo aliases for this. I'm reluctant to add another dependency to this project, such as `cargo-xtask`, which is also why I'm not crazy about using `nix-shell`, @kaichaosun .

Makefile
```diff
@@ -0,0 +1,14 @@
...
+	SKIP_WASM_BUILD= cargo check
+
+test:
+	SKIP_WASM_BUILD= cargo test --all
+
+run:
+	WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo run --release -- --dev --tmp
+
+build:
+	WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo build --release
```
athei, line 14:
```suggestion
.PHONY: init
init:
	./scripts/init.sh

.PHONY: check
check:
	SKIP_WASM_BUILD=1 cargo check

.PHONY: test
test:
	SKIP_WASM_BUILD=1 cargo test --all

.PHONY: run
run:
	WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo run --release -- --dev --tmp

.PHONY: build
build:
	WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo build --release
```

* Those are all [phony targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html)
* In my totally personal opinion ending a declaration with a `=` looks like a syntax error.

wheresaddie (approved):

wheresaddie:
just say no to cargo aliases

PR #115, opened 2020-12-05
Add .editorconfig and format code style

chenzongxiong:


.editorconfig
```diff
@@ -0,0 +1,22 @@
+root = true
+[*.rs]
+indent_style=tab
+indent_size=tab
+tab_width=4
+end_of_line=lf
+charset=utf-8
+trim_trailing_whitespace=true
+max_line_length=120
```
danforbes, line 9: I double-checked with others on this. I think we should enforce a maximum line length of 100-characters.

```suggestion
max_line_length=100
```

danforbes (changes requested):
Thank you so much! This is a great contribution :pray:

.editorconfig
chenzongxiong: From substrate documentation: 
```
Lines should be longer than 100 characters long only in exceptional circumstances and certainly no longer 
than 120. For this purpose, tabs are considered 4 characters wide.
```

Should I fix the line length long than 100?

danforbes: I think that would be the right course of action in this case. I'm happy to help with that or answer any questions you have along the way. You may want to look into using a tool like `cargo fmt` to help do this for you.

danforbes (approved):
Thank you so much!

PR #118, opened 2020-12-17
Init script should default to the correct nightly

danforbes:
cc: @NukeManDan

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -7,6 +7,7 @@ echo "*** Initializing WASM build environment"
 if [ -z $CI_PROJECT_NAME ] ; then
     rustup update nightly
     rustup update nightly-2020-10-05
+    rustup default nightly-2020-10-05
     rustup update stable
 fi
 
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -7,7 +7,6 @@ echo "*** Initializing WASM build environment"
 if [ -z $CI_PROJECT_NAME ] ; then
     rustup update nightly
     rustup update nightly-2020-10-05
-    rustup default nightly-2020-10-05
     rustup update stable
 fi
 
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,8 @@ test:
 
 .PHONY: run
 run:
-	WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo run --release -- --dev --tmp
+	WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo +nightly-2020-10-05 run --release -- --dev --tmp
 
 .PHONY: build
 build:
-	WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo build --release
+	WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo +nightly-2020-10-05 build --release


scripts/init.sh
```diff
@@ -7,6 +7,7 @@ echo "*** Initializing WASM build environment"
 if [ -z $CI_PROJECT_NAME ] ; then
     rustup update nightly
     rustup update nightly-2020-10-05
+    rustup default nightly-2020-10-05
```
riusricardo, line 10: No, don't change the standard config in the background.
Better to use the `+nightly` way.

danforbes: Yes that makes sense but didn't @athei think that was not the right approach to take?

NukeManDan:
https://github.com/rust-lang/cargo/issues/8989

This is giving an error for me. ☝🏼

scripts/init.sh
athei: I would never advocate for a script changing my config.

athei:
> [rust-lang/cargo#8989](https://github.com/rust-lang/cargo/issues/8989)
> 
> This is giving an error for me. ☝🏼
> 
> attempting to change for all OSes, so this works in the build file:
> `WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo +nightly-2020-10-05 build --release`

You can even drop the `+nightly-2020-10-05` as it is only the runtime that needs to be build with a specific nightly (or with nightly at all).

NukeManDan:
@athei everyone I know must set the default toolchain to this nightly to build. What we are trying to do is avoid `rustup default nightly-2020-10-05`

I _think_ that this flag is required if not default. I can do further testing if needed, but AFAIK without default set, builds fail.

scripts/init.sh
danforbes: No, that's not what I was trying to say, @athei - just that you had specifically advocate against the `+nightly...` approach in favor of the `WASM_BUILD_TOOLCHAIN...` approach.

danforbes:
@athei apologies as you may not have all the context behind this PR. I agree that it's probably not something we should merge, however, it was created because a developer was not able to build the Node Template until executing this command `rustup default nightly-2020-10-05`, as suggested by @xlc

xlc:
You absolute have to 

> > [rust-lang/cargo#8989](https://github.com/rust-lang/cargo/issues/8989)
> > This is giving an error for me. ☝🏼
> > attempting to change for all OSes, so this works in the build file:
> > `WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo +nightly-2020-10-05 build --release`
> 
> You can even drop the `+nightly-2020-10-05` as it is only the runtime that needs to be build with a specific nightly (or with nightly at all).

Unfortunately that's not the case anymore. If you don't like this approach (which isn't idea), please get https://github.com/paritytech/substrate/pull/7686 released.

athei:
Okay sorry then I don't get it. My latest information is that the **client** can be compiled with stable rust and it is only the **runtime** which needs nightly. And yes, because of a regression in rust nightly **and** the substrate fixes for that being unreleased it must be a specific nightly.

That said, setting the compiler version for the **runtime** is done via `WASM_BUILD_TOOLCHAIN` because that is picked up by the `build.rs` file. However, setting `+nightly-2020-10-05` will also compile the **client** with that nightly toolchain while it would compile perfectly fine with stable (which should be the default).

Is this information wrong or is there some additional stuff that applied to the node-template specifically that I don't know about?

xlc:
The native runtime cannot be build with a later nightly version so if people already default to a later nightly version, the build will fail. I think the latest stable is also not supported as well (needs a confirm).

NukeManDan:
Related: https://github.com/paritytech/substrate/issues/7150

I also have logged this issue: https://github.com/rust-lang/cargo/issues/8989

AFAIK builds of the template don't work with stable or nightly past 10/5/2020

tkimmm:
Could explore the possibilty of using rust-toolchain file in the root directory, less intrusive than nominating a default chain.

Couldn't locate the official documentation but essentially operates similiar to pyenv / nvm by creating a `rust-toolchain` file (no extension) in the root dir and placing the toolchain name within the file.

Active toolchain is then overridden.

danforbes:
Closing with #126

PR #125, opened 2021-01-12
bump to substrate 2.0.1 deps. No longer need older nightly toolchain!

NukeManDan:
### > Unbreak my build edition 🤞🏼 

---

Built the node template from scratch w/ bumped version & most recent toolchain:

```
$ rustup show
Default host: x86_64-unknown-linux-gnu
rustup home:  xxx

installed toolchains
--------------------

nightly-x86_64-unknown-linux-gnu

installed targets for active toolchain
--------------------------------------

wasm32-unknown-unknown
x86_64-unknown-linux-gnu

active toolchain
----------------

rustc 1.49.0 (e1884a8e3 2020-12-29)

```

Get the same error that v2.0.1 I thought was meant to fix:
```rust
error[E0282]: type annotations needed
let accuracy = P::ACCURACY.saturated_into();
```

also attempted to build with 

```
cargo +nightly build --release
```

Same issue. 

---
### NOTE: `wasm-builder-runner` depricated

https://crates.io/crates/substrate-wasm-builder-runner lists as depricated, and is used in the `build.rs` file for our runtime. Could this be part of the issue? (no 2.0.1 tag, v3 is deprecated)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1241,9 +1241,9 @@ dependencies = [
 
 [[package]]
 name = "frame-benchmarking"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "66a5e3fe43568300fdca1c1bfd45ea463a12cca8fbe6172a4f6d58cd54e3fbcc"
+checksum = "52b94f3dd88fd10ea9eb8bf6731ee22787deeb620c158766061ec1fc54170cda"
 dependencies = [
  "frame-support",
  "frame-system",
@@ -1279,9 +1279,9 @@ dependencies = [
 
 [[package]]
 name = "frame-executive"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c843800f05a7ad4653bc0db53a15e3d9bdd1cf14103e15c29e8aca200dbb1188"
+checksum = "cb0c4bcb8c34936657327c73c8a191159c998df425620bf4639737486d4bf5c0"
 dependencies = [
  "frame-support",
  "frame-system",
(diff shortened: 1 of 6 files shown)


NukeManDan:
The CI failed too 😭 

I retried with 
```
WASM_BUILD_TOOLCHAIN=nightly-2020-10-05 cargo build --release
```
after reverting to the old `build init` installing the 2020-10-05 nightly toolchain **but not setting rustup to this as a default** it still fails.

j-elmer123:
Hi @NukeManDan 

I'm new to rust. Is rust has virtual env (https://pypi.org/project/virtualenvwrapper/) like python?
If rust has that kind of feature, I think that you only need to update the `requirement` files and README.md

So, when others try the tutorial, they only to setup the virtual env, activate it, and run the project from that environment

NukeManDan:
@j-elmer123 thanks for the reference and resource! I am not much of a python dev :)

Rust though does not have a virtual environment, it is a [_compiled_ language vs an _interpreted_](https://www.lifewire.com/compiled-language-2184210) one like python. We do have docker functionality here in this repo, but I have not used it yet, and aim for this PR to fix the native build on any machine.

NukeManDan:
## Now builds and runs! 🎉 

```
active toolchain
----------------

rustc 1.49.0 (e1884a8e3 2020-12-29)
```

using `cargo build --release`

---

**NOTE:** serde 1.0.118 is required for upstream dependencies, so even with the manefest `Cargo.toml` file using 118, `multiaddr` throws an error:
```Compiling wasm-bindgen-futures v0.4.19
error[E0603]: module `export` is private
...
parity-multiaddr-0.9.6/src/onion_addr.rs:3:12
...
`export` is defined here
...
 serde-1.0.119/src/lib.rs:275:5
 ```
you *must* use cargo to update these deps (not by hand!) using `cargo update -p serde --precise 1.0.118` (otherwise your edits to the lock file get scrapped, checkout the diff on the lock file for details after running)

NukeManDan:
@danforbes the CI now passes! So we are good to review and merge, unless you wanted to use your results from the node template release tool in the substrate source?

danforbes:
Thank you for your work, @NukeManDan 🙏🏻 I am going to close this in favor of #126

Issue #127, opened 2021-01-14
Do not use depricated substrate-wasm-builder-runner


NukeManDan:
**Motivation**

This dependence is deprecated: https://github.com/substrate-developer-hub/substrate-node-template/blob/4d97032c11b8c65936e53a61607d4522d45a29ea/runtime/Cargo.toml#L14

**Suggested Solution**

Basti (`@bkchr`) suggests looking at substrate's master node implimentation to update to the new method: https://crates.parity.io/substrate_wasm_builder/index.html
See the docs here for details on how to migrate.

**Alternatives**

leave as is at our own risk ☠️ 

**Additional Information**

we should scan our dependencies for deprecations like this.... I am not sure how to automate it other than a `yank` from crates.io for dependencies, so you get  a warning when building. There may be cargo extensions to help...

danforbes:
I wonder if we should include this in #126?

NukeManDan:
Solved in #126

PR #128, opened 2021-01-14
Wasm builder upgrade

NukeManDan:
Solves #127 

I also ran a `cargo update` that bumped versions. This is presently based on #126 , if needed we can rebase to master. It's a simple patch here to migrate to the new builder.

based on [polkadot build.rs](https://github.com/paritytech/polkadot/blob/305d9414132ab2fb96f1b50cfbdb947f6be74793/runtime/polkadot/build.rs)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -609,6 +609,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "build-helper"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bdce191bf3fa4995ce948c8c83b4640a1745457a149e73c6db75b4ffe36aad5f"
+dependencies = [
+ "semver 0.6.0",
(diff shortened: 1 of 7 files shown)


NukeManDan:
Also of note: I have added the --release flag in our make file to test and check.

This is slightly streamlined, as the build times are faster after any of these commands are run once. Happy to remove this though, as I don't have hard evidence that it's an improvement.

NukeManDan:
:tada: Confirmed to build and run on my box:

```
installed targets for active toolchain
--------------------------------------

wasm32-unknown-unknown
x86_64-unknown-linux-gnu

active toolchain
----------------

rustc 1.49.0 (e1884a8e3 2020-12-29)
```
using `make build` and the included make file here

jimmychu0807:
@NukeManDan see if you think it is fair to close this PR given it is merged in #126

PR #130, opened 2021-01-15
Patch Multiaddress Into Substrate Node Template

shawntabrizi:
This PR is a quality of life improvement for developers and Jaco who manages custom types.

With this PR, we are patching in the Multiaddress update from Substrate master that was not included in v2.0.1.

This will make it so that the Polkadot JS Apps can by default use the Multiaddress format, which is what we will be using across Polkadot, Kusama, and many other parachain teams which have already switched.

When the node template is updated again to the latest master, this file can simply be deleted and directly replaced with what is in `sp_runtime::multiadress`.

cc @jacogr

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -94,9 +94,9 @@ pub mod opaque {
 pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("node-template"),
 	impl_name: create_runtime_str!("node-template"),
-	authoring_version: 100,
+	authoring_version: 1,
 	spec_version: 100,
-	impl_version: 100,
+	impl_version: 1,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 100,
 };
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -13,7 +13,7 @@ use sp_runtime::{
 	transaction_validity::{TransactionValidity, TransactionSource},
 };
 use sp_runtime::traits::{
-	BlakeTwo256, Block as BlockT, IdentityLookup, Verify, IdentifyAccount, NumberFor, Saturating,
+	BlakeTwo256, Block as BlockT, Verify, IdentifyAccount, NumberFor, Saturating,
 };
 use sp_api::impl_runtime_apis;
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
@@ -141,7 +141,7 @@ impl frame_system::Trait for Runtime {
 	/// The aggregated dispatch type that is available for extrinsics.
 	type Call = Call;
 	/// The lookup mechanism to get account ID from whatever is passed in dispatchers.
-	type Lookup = IdentityLookup<AccountId>;
+	type Lookup = multiaddress::AccountIdLookup<AccountId, ()>;
 	/// The index type for storing how many extrinsics an account has signed.
 	type Index = Index;
 	/// The index type for blocks.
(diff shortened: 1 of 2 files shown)


riusricardo (commented):
Thanks @shawntabrizi LGTM :)

If I recall correctly @jacogr also suggested to pump the template version (something like 100) so he can distinguish between different templates.

jacogr:
Yes, it goes much further than the UI, but also ensuring the API just works out of the box. 

We know we are driving everybody in the direction of MultiAddress now, so imho the earlier we get the ball rolling the better. Since this is a starting point of a lot of people, we really want to align with where we want to be.

jimmychu0807:
> Thanks @shawntabrizi LGTM :)
> 
> If I recall correctly @jacogr also suggested to pump the template version (something like 100) so he can distinguish between different templates.

@riusricardo Do you mean in the `runtime/src/lib.rs`, one of the version values below to be set to 101? 

```rust
pub const VERSION: RuntimeVersion = RuntimeVersion {
	spec_name: create_runtime_str!("node-template"),
	impl_name: create_runtime_str!("node-template"),
	authoring_version: 1,
	spec_version: 1,
	impl_version: 1,
	apis: RUNTIME_API_VERSIONS,
	transaction_version: 1,
};
```

Or the node-template version in `Cargo.toml` to be set to `2.0.101`?

riusricardo:
@jimmychu0807 Yes,  I'm talking about the spec_version. We need to start from a higher number. 0-99 will be de old and >100 will be the multiaddress templates.

jimmychu0807:
Make [this commit](https://github.com/substrate-developer-hub/substrate-node-template/pull/130/commits/b5a2f776ae6b06d7685525a7706169bf4d6c8ceb).

See if this looks right to you @riusricardo @jacogr

runtime/src/lib.rs
```diff
@@ -94,11 +94,11 @@ pub mod opaque {
...
 	impl_name: create_runtime_str!("node-template"),
-	authoring_version: 1,
-	spec_version: 1,
-	impl_version: 1,
+	authoring_version: 100,
+	spec_version: 100,
+	impl_version: 100,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 100,
```
jacogr, line 101: Just needs spec_version here, the rest can be 1.

danforbes:
Should we make a companion PR to `bin/node-template` from Substrate?

shawntabrizi:
@danforbes no, Substrate master already has this enabled for node template in the correct way

runtime/src/lib.rs
```diff
@@ -94,11 +94,11 @@ pub mod opaque {
 pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("node-template"),
 	impl_name: create_runtime_str!("node-template"),
-	authoring_version: 1,
-	spec_version: 1,
-	impl_version: 1,
+	authoring_version: 100,
+	spec_version: 100,
+	impl_version: 100,
```
shawntabrizi, line 99:
```suggestion
	authoring_version: 1,
	spec_version: 100,
	impl_version: 1,
```

runtime/src/lib.rs
shawntabrizi, line 101:
```suggestion
	transaction_version: 1,
```

shawntabrizi: done

jacogr:
We may want to bump Substrate master spec_version to 100 as well to align here.

jimmychu0807 (approved):

jimmychu0807:
Thank you @shawntabrizi  @jacogr 

@danforbes The only question left is whether we should tag this commit and how. If there is no major repercussion, I prefer moving the tag and call this `v2.0.1`. If you come up with any valid reason, then make this `v2.0.1+1` as you mentioned. Your call.

PR #131, opened 2021-01-16
rename-node-runtime script completed

jimmychu0807:
Hi devhub team, 

I created a js script that can rename the node and runtime. The reason is that I have heard from ppl building pallet using node-template that it is hard to rename them as it requires one to modify multiple place. This script is meant to do this. 

## Usage

```bash
# node-name is `my-app`, and runtime is `my-app-runtime`
./rename-node-runtime -n my-app  

# node-name is `my-app`, and runtime name is `my-runtime`
./rename-node-runtime -n my-app -r my-runtime 

# with custom substrate path
./rename-node-runtime -n my-app -r my-runtime -p ~/my-substrate-path
```

You can check out the branch and try it out.

## Feedback Needed

1. Do you think this scripts is useful, or kind of redundant? And that ppl should be able to do this themselves manually?
2. If this script is useful, where should we put it? In `substrate-node-template` repo, or a `node-template-utils` repo?
3. Should I do a `rename-pallet` script as well?

@danforbes @NukeManDan @riusricardo @kaichaosun would like to hear your feedback. Thanks.


NukeManDan:
Yeah IMHO this is actually a bit _more_ confusing as you need to understand the script, and also understand then what it is doing within your code. So it actually make the user's job harder 😅 

I would think better devhub docs to point this out would be a better path to mitigate this, and instead have users focus on what matters: understanding their substrate-based code deeply.

jimmychu0807:
let close it then.

PR #135, opened 2021-01-28
fix cargo clippy warnning

DaviRain-Su:


riusricardo:
Thanks for the PR, most of this changes will get on the next template release as they are already fixed in master.

PR #141, opened 2021-02-12
Upgrade to v3.0.0

danforbes:
This PR also removes the `init.sh` script and the `Makefile`.


shawntabrizi (approved):
Why remove init.sh?

codywalker77 (approved):

danforbes:
> Why remove init.sh? It's pretty generic and useful to have I think

I think in general we've decided to favor "declarative configuration" as opposed to "hiding" things behind init scripts, Makefiles, etc.

riusricardo (approved):

PR #143, opened 2021-02-23
Rely on debug compilation

jeluard:


apopiak:
You're sure this won't lead to consens us issues? That's what I heard in the past why we use `--release`

jeluard:
That's the template for playground workshops, what do you want to consensus with? :D

apopiak:
Well in that case I guess I mean just block production.
But vever mind, if it works it works.

PR #160, opened 2021-03-28
The latest Substrate Node Template v3 compiled from Substrate

jimmychu0807:
- Using latest `#[pallet::*]` attribute macro.
- With benchmarking for pallet-template
- Will be tagged `v3.0.0+1`

@shawntabrizi  @NukeManDan See if you have any concern on this PR. Thanks


Cargo.toml
```diff
@@ -1,9 +1,8 @@
-[profile.release]
-panic = 'unwind'
-
 [workspace]
 members = [
     'node',
-    'pallets/*',
```
NukeManDan, line 7: Why remove the wildcard here? Thus needs to be reintroduced in the tutorials as a note if its required

node/src/chain_spec.rs
```diff
@@ -39,7 +39,7 @@ pub fn authority_keys_from_seed(s: &str) -> (AuraId, GrandpaId) {
 }
 
 pub fn development_config() -> Result<ChainSpec, String> {
-	let wasm_binary = WASM_BINARY.ok_or_else(|| "Development wasm binary not available".to_string())?;
```
NukeManDan, line 42: Why remove the binary note? Wasm tooling can be available, but the binary missing. No?

node/src/chain_spec.rs
```diff
@@ -78,7 +78,7 @@ pub fn development_config() -> Result<ChainSpec, String> {
 }
 
 pub fn local_testnet_config() -> Result<ChainSpec, String> {
-	let wasm_binary = WASM_BINARY.ok_or_else(|| "Development wasm binary not available".to_string())?;
```
NukeManDan, line 81: Same

node/src/chain_spec.rs
```diff
@@ -134,24 +134,24 @@ fn testnet_genesis(
 	_enable_println: bool,
 ) -> GenesisConfig {
 	GenesisConfig {
-		frame_system: Some(SystemConfig {
```
NukeManDan, line 137: Is there precedence removing the option? These fields are all required now?

pallets/template/Cargo.toml
```diff
@@ -5,22 +5,18 @@ edition = '2018'
...
-# alias "parity-scale-code" to "codec"
-[dependencies.codec]
-default-features = false
-features = ['derive']
-package = 'parity-scale-codec'
-version = '2.0.0'
-
 [dependencies]
-frame-support = { default-features = false, version = '3.0.0' }
+codec = { default-features = false, features = ['derive'], package = 'parity-scale-codec', version = '2.0.0' }
```
NukeManDan, line 16: Should this be 3?

runtime/Cargo.toml
```diff
@@ -13,19 +13,10 @@ targets = ['x86_64-unknown-linux-gnu']
...
 
-# alias "parity-scale-code" to "codec"
-[dependencies.codec]
-default-features = false
-features = ['derive']
-package = 'parity-scale-codec'
-version = '2.0.0'
-
 [dependencies]
+codec = { default-features = false, features = ['derive'], package = 'parity-scale-codec', version = '2.0.0' }
```
NukeManDan, line 17: V3?

runtime/src/lib.rs
```diff
@@ -96,14 +96,14 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("node-template"),
 	impl_name: create_runtime_str!("node-template"),
```
NukeManDan, line 97: @shawntabrizi is there a specific reason not to start at 1? Or 0 for that matter?

runtime/src/lib.rs
```diff
@@ -96,14 +96,14 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
...
+	// TODO: adding reason on setting spec_version to 100.
 	spec_version: 100,
 	impl_version: 1,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
 };
 
-
-/// This determines the average expected block time that we are targeting.
+/// This determines the average expected block time that we are targetting.
```
NukeManDan, line 106: Targeting ;)

runtime/src/lib.rs
```diff
@@ -324,7 +324,7 @@ impl_runtime_apis! {
 		}
 
 		fn execute_block(block: Block) {
-			Executive::execute_block(block)
+			Executive::execute_block(block);
```
NukeManDan, line 327: The ; stops this fn from returning a value, right? Why do we do this here?

runtime/src/lib.rs
```diff
@@ -380,8 +379,8 @@ impl_runtime_apis! {
 	}
 
 	impl sp_consensus_aura::AuraApi<Block, AuraId> for Runtime {
-		fn slot_duration() -> u64 {
-			Aura::slot_duration()
+		fn slot_duration() -> sp_consensus_aura::SlotDuration {
```
NukeManDan, line 382: Was this a bug previously? It seems you have introduced a conversion that previously wasn't there?

NukeManDan (commented):
Hope these comments help! They for sure help clarify for me to help me understand. 🙇🏼 

Also with telemetry, are we documenting how to use this somewhere? As well as benchmarking?

jimmychu0807:
@NukeManDan This Node Template PR is made from upstream, the Substrate repo. A simple reason about changes made above is because the upstream doesn't have these. But we can incorporate the goods back to upstream.

This is a related PR I made in Substrate (under review): 
https://github.com/paritytech/substrate/pull/8473

With the node template release process documented here:
https://github.com/paritytech/substrate/pull/8473/files#diff-196935745b83692bbe4ec984962de8306805e6721a87e91288c9547ed7d91a67

Cargo.toml
jimmychu0807: Sure, let me add back. Just that this auto-generated file doesn't use `*` originally.

node/src/chain_spec.rs
jimmychu0807: This is just a change in wording. The error message is just changed from `Development wasm binary not available` to `Development wasm not available`. This shouldn't be a big deal.

node/src/chain_spec.rs
jimmychu0807: ditto

jimmychu0807 (commented):
@NukeManDan But I just realized a key caveat, I have to revert some latest changes because those changes depend on package features that are not published to crate.io yet.

Copied over from [substrate node process release doc](https://github.com/paritytech/substrate/pull/8473/files#diff-196935745b83692bbe4ec984962de8306805e6721a87e91288c9547ed7d91a67):

> Notice there are chances the code in Substrate Node Template works with the linked Substrate git
	commit, but not with published packages due to the latest unpublished (yet) features. In this case,
	resume that section of Node Template to its previous version to ensure Node Template builds.

jimmychu0807:
@NukeManDan thanks for taking a look and reviewing this PR 🙏

NukeManDan:
Lgtm minus the spec version justification.

Also with telemetry, are we documenting how to use this somewhere? As well as benchmarking?

NukeManDan (approved):

jimmychu0807:
**telemetry**: we should add a todo on that.
**benchmarking**: I added enough inline code for benchmarking in node-template, and also the [Runtime benchmarking kb article](https://substrate.dev/docs/en/knowledgebase/runtime/benchmarking).

NukeManDan:
Ok so a new issue on the devhub for telem?

and there should be a check that nothing depends on FRAME v1 on devhub, else the new sintax will confuse people. Most notably the create a pallet tutorial. Should be prioritized now that this is merged.

Issue #166, opened 2021-04-04
Benchmarking is broken


fzzr-:
Actual for version (tag) `v3.0.0+1`.

It's impossible to build node with benchmarking feature because of:
- [`use frame_system_benchmarking::Pallet`](https://github.com/substrate-developer-hub/substrate-node-template/blob/v3.0.0%2B1/runtime/src/lib.rs#L463)
- [there's `Module` only, not `Pallet`](https://github.com/paritytech/substrate/blob/v3.0.0/frame/system/benchmarking/src/lib.rs#L37)

So, just replace that line with `use frame_system_benchmarking::Module as SystemBench;` please :)

NukeManDan:
@jimmychu0807 would you mind fixing and testing this one? Happy to take it on if you don't want it.

fzzr-:
Okay. That was a little more "complex" than I said earlier. I've done.

jimmychu0807:
@NukeManDan will look into this.

PR #167, opened 2021-04-05
Fix broken benchmarking

fzzr-:
This fixes [issue #166](https://github.com/substrate-developer-hub/substrate-node-template/issues/166).

- fix `frame_system_benchmarking::Module` usage;
- add missed sp-std dependency needed for `benchmarks!` macro;
- add missed sub-features for `runtime-benchmarks` feature;
- add build CI-check with benchmarking feature


fzzr-:
That was a little more "complex" than I said earlier. I've done.

.github/workflows/check.yml
```diff
@@ -37,3 +37,8 @@ jobs:
       - name: Check Build
         run: |
           SKIP_WASM_BUILD=1 cargo check --release
+
+      - name: Check Build for Benchmarking
+        run: >
+          pushd node &&
+          cargo check --features=runtime-benchmarks --release
```
thiolliere, line 44: I don't think we need release flag for checking the build

pallets/template/Cargo.toml
```diff
@@ -18,6 +18,8 @@ frame-system = { default-features = false, version = '3.0.0' }
 frame-support = { default-features = false, version = '3.0.0' }
 frame-benchmarking = { default-features = false, optional = true, version = '3.1.0' }
 
+sp-std = { default-features = false, optional = true, version = '3.0.0' }
```
thiolliere, line 21: optional is not strictly necessary IMO, in importing sp-std is good to do anyway to have easy access on types.

thiolliere (approved):
looks good to me.

NOTE that benchmark macro has been fixed in substrate master, we could release frame-benchmarking v3.2.0 with the fix.

.github/workflows/check.yml
NukeManDan: agreed, I started using this assuming that the same build artifacts would be used for `check` and `build`, but from what I have seen experimenting is that they do not do this 😭  , not needed, but also not a bad idea?

fzzr-: As I see in [CI checks](https://github.com/substrate-developer-hub/substrate-node-template/pull/167/checks) there's some artefacts used for second build because only 2 minutes for it.
Screenshot 2021-04-12 at 22 33 07
So, second build process is without downloading and compiling crates where no benchmarking-feature.

pallets/template/Cargo.toml
fzzr-: fixed [there](https://github.com/substrate-developer-hub/substrate-node-template/pull/167/commits/2f2b7316108dcc0c256f3104f458c4732425ab02)

.github/workflows/check.yml
fzzr-: fixed [there](https://github.com/substrate-developer-hub/substrate-node-template/pull/167/commits/2f2b7316108dcc0c256f3104f458c4732425ab02) but look at [times there](https://github.com/substrate-developer-hub/substrate-node-template/runs/2327248731).

fzzr-: Well, I'm sure that's not good, so I'd like to return `release` back if you don't mind.
Screenshot 2021-04-12 at 23 06 07

thiolliere: ah indeed, then ok release flag is better.

NukeManDan: Thanks for investigating! Good yo know

jimmychu0807:
Indeed very cool @fzzr-. Thanks for the fix.

NukeManDan:
> NOTE that benchmark macro has been fixed in substrate master, we could release frame-benchmarking v3.2.0 with the fix.
@thiolliere is there a tracking issue? If so, we can point to that and update the template when it lands, not just here though, in the upstream node template. 

Is that already in the works?

thiolliere:
AFAIK the benchmark macro has been fixed in substrate master, I'm not sure there is a test though.

PR #170, opened 2021-04-19
Add nix build

Anderssorby:
This PR adds a more plug and play experience for starting development with substrate by using nix.


al3mart:
Awesome! Thanks @Anderssorby

nix/rust.nix
```diff
@@ -0,0 +1,10 @@
+{ sources ? import ./sources.nix }:
+let
+  pkgs =
+    import sources.nixpkgs { overlays = [ (import sources.nixpkgs-mozilla) ]; };
+  channel = "nightly";
+  date = "2021-03-18";
```
NukeManDan, line 6: this adds a new thing we must track and update for us. I would be inclined to let this go to latest, with a note for users to set this if errors crop up.

Anderssorby: That is sort of against the nix philosophy though. Since you should always get the same result for the same input. So updates should always be explicit, but it is probably possible to do. It is also only possible to use the versions from the last time `niv update` was run. So 'latest' would always be relative to sources.json

bkchr (commented):
Having just a simple `shell.nix` would be better IMHO.

Anderssorby:
@bkchr That will be more feasible once flakes.nix becomes stable. For now a stand alone shell.nix is not very practical for development since it would be dependent on the nix channels and generally not very flexible. Niv solves that by having a per project definition which leads to some boiler plate, but also a nicer experience.

bkchr:
https://github.com/paritytech/substrate/blob/master/shell.nix

We have the following shell.nix file in substrate and that is not really dependent on the nix channel 🤷

Anderssorby:
I guess it is more suitable for you with a single file, but when you are using `<nixpkgs>` (with a mozilla overlay) you are dependent on the nix-channel. Often this is not a problem, but updates to software on the channel may lead to broken builds. It is also sensitive to the version of nixpkgs or nixos. So what you can do is pin it to ensure that it will behave the same way even after many years which is what `niv` simplifies a lot. So the question is: do you want to pin it with niv or not?

Anderssorby:
@bkchr I changed to using a stand alone `shell.nix` with pinned nixpkgs based on the substrate one. So this one should be much more minimal.

bkchr (approved):

NukeManDan:
@bkchr should this be also incorporated into the substrate node template in the substrate core repo? Or is it only useful in context of a stand alone template like this?

PR #183, opened 2021-05-13
Added instructions for custom-types

jimmychu0807:
@sacha-l @NukeManDan 

Adding instructions on the custom-types. Otherwise the Polkadot-JS Apps cannot decode the account data correctly from Substrate runtime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ A fresh FRAME-based [Substrate](https://www.substrate.io/) node, ready for hacki
 
 ## Getting Started
 
-Follow the steps below to get started with the Node Template, or get it up and running right from your browser 
+Follow the steps below to get started with the Node Template, or get it up and running right from your browser
 in just a few clicks using [Playground](https://playground.substrate.dev/) :hammer_and_wrench:
 
 ### Using Nix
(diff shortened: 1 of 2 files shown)


shawntabrizi:
@jimmychu0807 why not update the node template?

NukeManDan:
Ricardo suggests that we start to version the template here, and bump as needed when things like this arise.

Is the right path to do a release via this guide? https://github.com/paritytech/substrate/blob/master/docs/node-template-release.md

If so, I can give it a shot, and make a 3.0.0+2 tag for it

jimmychu0807:
@shawntabrizi Currently all dependent packages in node template are retrieved from Cargo crate registry, and the latest triple ref count in `frame-system` has not been published there. If we want to update the node-template, we would have the dependent packages in `Cargo.toml` pointing back to Substrate github repo with tag (`monthly-2021-05` would work). Is this approach good?

jimmychu0807:
> Ricardo suggests that we start to version the template runtime spec_version here I think, and bump as needed when things like this arise.

I think this will be good. Will need the coordination between core devs & Jaco on this.

davidrhodus:
Could this type just be added into Polkadot-JS App?

jimmychu0807:
> Could this type just be added into Polkadot-JS App?

Yes, this is the instruction added in this PR.

davidrhodus:
> @davidrhodus
> 
> > Could this type just be added into Polkadot-JS App?
> 
> Yes, this is the instructions added in this PR. Is it not obvious? 🤔

Your instructions are extremely helpful as they have unblocked users at the moment.

I am asking about changing the front end https://github.com/polkadot-js/apps so that a users doesn't have to manually add in the type to make the front end work with node-template work.

PR #186, opened 2021-05-17
Node template generated from Substrate repo `monthly-2021-05` tag

jimmychu0807:
@shawntabrizi @riusricardo @apopiak @sacha-l @NukeManDan 

This is the node template generated from Substrate repo [`monthly-2021-05` tag](https://github.com/paritytech/substrate/tree/monthly-2021-05). There are quite a lot of code changes, and I would like to merge it to master branch (and tagged as `v3.0.0+2`).  

Only concern is this node template all refer back to Substrate github repo instead of Cargo crate. But I actually prefer this, and this is closer to the latest development status.

Related: https://github.com/substrate-developer-hub/substrate-node-template/pull/183#issuecomment-840937155


apopiak:
Why not mirror the structure in Substrate and have a monthly tag? You can still publish it as a release
I think it would be very confusing to have such significant changes in the node template between `v3.0.0+1` and `v3.0.0+2`

NukeManDan:
I am for a monthly tag, but only actually updating and pulling in changes for the template as needed, when things break (as they did for #183 ), and keeping the references to what is working for the devhub (as a base and for all solutions that are derived from that base commit) as stable as possible. Most people use this template, and if we have many users that have many different versions they are trying to troubleshoot, it could become a nightmare. 

So the flow I would prefer would be to wait for a breaking change or tagged semver release of substrate before updating to a new template to fix and propagate this to the devhub for all tutorials (and solutions!!) in one PR.

NukeManDan:
Possibly using the tagging format as: `v<semver>-<yyyy-mm>` for the tags? so `v3.0.0-2021-05`  would be the tag for this PR once merged as an example.

node/src/service.rs
```diff
@@ -110,7 +136,7 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
 		select_chain,
 		transaction_pool,
 		inherent_data_providers,
-		other: (block_import, grandpa_link),
+		other: (block_import, grandpa_link, mut telemetry),
```
NukeManDan, line 139: does this need to be mut, while the last one doesn't? line 117

node/src/service.rs
```diff
@@ -327,6 +383,7 @@ pub fn new_light(mut config: Configuration) -> Result<TaskManager, ServiceError>
 		network,
 		network_status_sinks,
 		system_rpc_tx,
+		telemetry: telemetry.as_mut(),
```
NukeManDan, line 386: same mut question.

runtime/src/lib.rs
```diff
@@ -365,7 +369,7 @@ impl_runtime_apis! {
 		}
 
 		fn random_seed() -> <Block as BlockT>::Hash {
-			RandomnessCollectiveFlip::random_seed()
+			RandomnessCollectiveFlip::random_seed().0
```
NukeManDan, line 372: Does this impact randomness at all? Why is the .0 needed here?

NukeManDan (approved):
Looks good, and compiles locally 😁 
I had more a few clarifications in comments here. Shouldn't be blockers.

jimmychu0807:
@NukeManDan referring to your questions, those code changes are not made by me, but made upstream in Substrate following `monthly-2021-05` tag. Refer to:

- [on `telemetry` becoming **mut**](https://github.com/paritytech/substrate/blame/monthly-2021-05/bin/node-template/node/src/service.rs#L139)
- [on `random_seed()` becoming `random_seed().0`](https://github.com/paritytech/substrate/blame/monthly-2021-05/bin/node-template/runtime/src/lib.rs#L372)

jimmychu0807:
So it seems the consensus is that we will release a node template in an on-demand basis, either when a semver version is released, or breaking changes/bug fixes are made and we will use the latest monthly tag version for release.

Will use `v3.0.0+monthly-2021-05` as the tag name. The suffix I will try to follow what Substrate tag is using.

PR #189, opened 2021-05-28
fix missing docker-compose volume source (./.local)

hamidra:
The docker-compose.yml file binds a volume from ./.local path but if the path does not exist it fails (e.g. a fresh clone from this repo will fail to run docker). Changes are made to make docker_run.sh to make the directory if it does not exist.


NukeManDan (approved):
Thanks for the improvement!

PR #210, opened 2021-07-02
`monthly-2021-07` release

jimmychu0807:
@NukeManDan @sacha-l 
node template based on Substrate `monthly-2021-07` tag

**Please tag the commit as `v3.0.0+monthly-2021-07` after merging the PR**.


sacha-l (approved):
LGTM!

riusricardo (approved):

Cargo.toml
```diff
@@ -1,9 +1,8 @@
-[profile.release]
-panic = 'unwind'
-
```
NukeManDan, line 3: Curious. Do you know the reason to do this @jimmychu0807 ? Seems like a good thing for most people to use, no? I think it gives much better debugging for panics

jimmychu0807: These two lines are still there, just being moved to another location in the file.

Issue #214, opened 2021-07-09
Add several storage example in substrate-node-template


branciard:
**Motivation**

I try to add basic HashMap in FRAME v2 syntax in the new substrate-node-template

https://substrate.dev/docs/en/knowledgebase/runtime/storage#declaring-storage-items

like 

```
#[pallet::storage]
pub(super) type SomeComplexValue<T> = StorageValue<_, T::AccountId, ValueQuery>;
```

But I have error : 
```
associated type `AccountId` not found
```
It certainly easy to fix, but I search in others examples and substrate code but do not find how to fix it.

**Suggested Solution**

add all storage examples described here https://substrate.dev/docs/en/knowledgebase/runtime/storage#declaring-storage-items
in the substrate-node-template lib.rs example

**Alternatives**
At least add commented storage example lines. For a copy/paste usage.

sacha-l:
Thank you @branciard for creating [this how-to guide](https://substrate.dev/substrate-how-to-guides/docs/pallet-design/storage-map) to address this issue. Closing this now.

PR #215, opened 2021-07-12
Add StorageMap Example

branciard:
Proposal for #214

### DIFF
--- a/pallets/template/src/lib.rs
+++ b/pallets/template/src/lib.rs
@@ -39,6 +39,11 @@ pub mod pallet {
 	// https://substrate.dev/docs/en/knowledgebase/runtime/storage#declaring-storage-items
 	pub type Something<T> = StorageValue<_, u32>;
 
+	// StorageMap Example
+	#[pallet::storage]
+	#[pallet::getter(fn when_last_something_done)]
+	pub type WhenLastSomethingDone<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, T::BlockNumber, ValueQuery>;
+
 	// Pallets use events to inform users when important changes are made.
 	// https://substrate.dev/docs/en/knowledgebase/runtime/events
 	#[pallet::event]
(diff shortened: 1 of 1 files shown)


NukeManDan:
I like where this is going! @sacha-l do you think this kind of thing might be a fit for https://github.com/substrate-developer-hub/substrate-how-to-guides/ instead of changing the template here? If we wanted to include this on the template, we would need to push it upstream: https://github.com/paritytech/substrate/tree/master/bin/node-template

sacha-l:
@NukeManDan not sure if this means to update our KB article, make a HTG or both. Thoughts?

NukeManDan:
Maybe both @sacha-l ? I would leave it up to your discretion here.

sacha-l:
The addition certainly gives beginner developers a clearer idea of the different storage items in Substrate. However it adds a little more fluff to the node template's pallet-template which I'm not sure is desirable from an educational perspective.  The tradeoffs I see:
- on the one hand it adds complexity, on the other gives folks a better idea of the storage items they could use (and how to use them)
-  it would imply that we write tests for all additional storage items, as well as Events and Errors, adding more fluff 

My view is: we want the Node Template to be useful but at the same time educational and not too complicated so that beginners can easily navigate it. For developers already familiar with the mechanics of a pallet, the Recipes and How-to Guides are the go to ressource. The template pallet is designed to be the most simplest pallet to be built upon.

What I suggest @branciard is perhaps you bring this issue over [here](https://github.com/substrate-developer-hub/substrate-how-to-guides/issues), proposing to make a How-to Guide for adding different storage items to a pallet. IMO the code you added would be a nice addition to the guides, showing how to use a `StorageMap` in some context, for e.g. a StorageMap that throws an error if a value is not some specific type specified by the dispatchable function (i.e. `u32`, `T::BlockNumber` ..).

Feel free to contribute with a PR directly by following [this template](https://substrate.dev/substrate-how-to-guides/docs/intro#template) :)

branciard:
Thank you for the feedback @NukeManDan and @sacha-l .
I initialize a PR guide [here](https://github.com/substrate-developer-hub/substrate-how-to-guides/pull/56) but as a real beginner newbie I do not  include Error treatment you propose as I didn't manage to do this. But I will be glad to see how it can be managed properly.

NukeManDan:
Thanks for the contribution @branciard 😀

PR #222, opened 2021-08-01
Update to substrate `monthly-2021-08`

NukeManDan:
Follows guide https://github.com/paritytech/substrate/blob/master/docs/node-template-release.md

Preserves README with edits for this template. 

I have updated the cargo files to include the monthly version correctly, and cleaned them up a bit for easier navigation.

Built with:
```bash
$ rustup show active-toolchain --verbose
rustc 1.54.0 (a178d0322 2021-07-26)

$ rustup +nightly show active-toolchain --verbose
nightly-x86_64-unknown-linux-gnu (overridden by +toolchain on the command line)
rustc 1.56.0-nightly (4e282795d 2021-07-31)
```

Tested to [work with the FE template](https://github.com/substrate-developer-hub/substrate-front-end-template/issues/188)

### DIFF
--- a/node/src/rpc.rs
+++ b/node/src/rpc.rs
@@ -9,10 +9,10 @@ use std::sync::Arc;
 
 use node_template_runtime::{opaque::Block, AccountId, Balance, Index};
 pub use sc_rpc_api::DenyUnsafe;
+use sc_transaction_pool_api::TransactionPool;
 use sp_api::ProvideRuntimeApi;
 use sp_block_builder::BlockBuilder;
 use sp_blockchain::{Error as BlockChainError, HeaderBackend, HeaderMetadata};
-use sp_transaction_pool::TransactionPool;
 
 /// Full client dependencies.
 pub struct FullDeps<C, P> {
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -32,7 +32,7 @@ pub fn new_partial(
 		FullClient,
 		FullBackend,
 		FullSelectChain,
-		sp_consensus::DefaultImportQueue<Block, FullClient>,
+		sc_consensus::DefaultImportQueue<Block, FullClient>,
 		sc_transaction_pool::FullPool<Block, FullClient>,
 		(
 			sc_finality_grandpa::GrandpaBlockImport<
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,7 +18,16 @@ version = "0.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e7a2e47a1fbe209ee101dd6d61285226744c6c8d3c21c8dc878ba6cb9f467f3a"
 dependencies = [
- "gimli",
+ "gimli 0.24.0",
+]
+
+[[package]]
+name = "addr2line"
+version = "0.16.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
+dependencies = [
+ "gimli 0.25.0",
 ]
 
 [[package]]
@@ -57,7 +66,7 @@ dependencies = [
  "aes",
  "block-cipher",
  "ghash",
- "subtle 2.4.0",
+ "subtle 2.4.1",
 ]
 
 [[package]]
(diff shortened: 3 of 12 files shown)


sacha-l (commented):
@NukeManDan - I just have a few things I'm unsure about here, thank you in advance for helping me better understand your proposed changes:
- why are we reverting back to using `rev` instead of `tag` in `node/Cargo.toml` and `runtime/Cargo.toml`? Shouldn't we be using the `latest` or `monthly-2021-08` tags instead?
- in `node/src/rpc.rs` you changed the source of `TransactionPool`, from `sc_transaction_pool` to `sc_transaction_pool_api`. Why? 
- in `node/src/service.rs`, `DefaultImportQueue` was updated too. Why? What implications does this have on its rustdoc [here](https://crates.parity.io/sc_consensus/import_queue/type.DefaultImportQueue.html) and [here](https://substrate.dev/rustdocs/latest/sp_consensus/import_queue/type.DefaultImportQueue.html) ?

NukeManDan:
> why are we reverting back to using `rev` instead of `tag` in `node/Cargo.toml` and `runtime/Cargo.toml`? Shouldn't we be using the `latest` or `monthly-2021-08` tags instead?

This is the output of the automated script, I am happy to manually update these. I was under the {seems false} presumtion this would be automated if the commit referenced has a tag associated. @jimmychu0807 you did the last one, did you need to manually update for the monthly tag?

> in `node/src/rpc.rs` you changed the source of `TransactionPool`, from `sc_transaction_pool` to `sc_transaction_pool_api`. Why?

> in `node/src/service.rs`, `DefaultImportQueue` was updated too. Why? What implications does this have on its rustdoc [here](https://crates.parity.io/sc_consensus/import_queue/type.DefaultImportQueue.html) and [here](https://substrate.dev/rustdocs/latest/sp_consensus/import_queue/type.DefaultImportQueue.html) ?


It's much better practice to comment on the file change to point directly 🙏🏼 I will do so presently.

node/src/rpc.rs
```diff
@@ -9,10 +9,10 @@ use std::sync::Arc;
 
 use node_template_runtime::{opaque::Block, AccountId, Balance, Index};
 pub use sc_rpc_api::DenyUnsafe;
+use sc_transaction_pool_api::TransactionPool;
```
NukeManDan, line 12: This was pulled in from upstream. You can look at the blame log here: https://github.com/paritytech/substrate/blame/5e1aab8192cd84b653e023db5ee83abda856438c/bin/node-template/node/src/rpc.rs#L12

node/src/service.rs
```diff
@@ -32,7 +32,7 @@ pub fn new_partial(
 		FullClient,
 		FullBackend,
 		FullSelectChain,
-		sp_consensus::DefaultImportQueue<Block, FullClient>,
+		sc_consensus::DefaultImportQueue<Block, FullClient>,
```
NukeManDan, line 35: https://github.com/paritytech/substrate/blame/5e1aab8192cd84b653e023db5ee83abda856438c/bin/node-template/node/src/service.rs#L35

NukeManDan: I think the rustdocs are built off of an earlier release that you reference? This should fix itself as soon as they are published on a commit younger than this blame.

sacha-l: Yes but the issue I'm raising is it's the latest tag, therefore should be updated right? Otherwise it adds confusion among versioning and crate libraries.

NukeManDan: What is the latest tag here? I updated to the monthly tag as you suggested already. Am I missing a different change? Please point in the code directly if so 🙏🏼

sacha-l (approved):
LGTM!

sacha-l:
What I was raising in the above comments is somewhat not entirely relevant to your PR. But to the broader problem that results in there being inconsistencies with what one should expect is the most updated version of the rustdocs and crates. The latest tag [here](https://substrate.dev/rustdocs/latest/sp_consensus/import_queue/type.DefaultImportQueue.html) refers to `sp_consensus` but the [Parity crates one](https://crates.parity.io/sc_consensus/import_queue/type.DefaultImportQueue.html) (i.e. what _should_ be the reference for latest) refers to `sc_consensus`.

jimmychu0807:
> But to the broader problem that results in there being inconsistencies with what one should expect is the most updated version of the rustdocs and crates. The latest tag [here](https://substrate.dev/rustdocs/latest/sp_consensus/import_queue/type.DefaultImportQueue.html) refers to `sp_consensus` but the [Parity crates one](https://crates.parity.io/sc_consensus/import_queue/type.DefaultImportQueue.html) (i.e. what _should_ be the reference for latest) refers to `sc_consensus`.

@sacha-l This issue arise because the current latest rustdocs is still at `monthly-2021-07`. Let me update the rustdocs to `monthly-2021-08` one and have `latest` point to it, then it should behave as expected.

PR #223, opened 2021-08-02
Add a matrix chat badge to the README

NukeManDan:
https://github.com/paritytech/substrate/pull/9476/ (now merged) - so the upstream substrate README includes this.


NukeManDan:
If we like this here (and upstream is substrate) we can include it in other devhub org READMEs as we see fit. Only bummer is the 3rd party API used for generating this

sacha-l (approved):
LGTM ! :)

PR #231, opened 2021-09-03
Kitties Update

jimmychu0807:
@sacha-l

I updated the Kitty Tutorial to be based on `v3.0.0+monthly-2021-08`. This will cause a lot of changes when looking at the code diff, but the major changes are mostly in `pallets/kitties/src/lib.rs`.

Major changes: 

- Simplified to use only 3 storage pieces.
- Using **BoundedVec** for storing what kitties are owned by which user. The length of the BoundedVec, `MaxKittyOwned`, is passed in from runtime.
- Having **Currency** type in Config trait, and is passed in from runtime. It pretty much do what `Balance` allowed before, but also support checking user [`free_balance()`](https://substrate.dev/rustdocs/latest/frame_support/traits/tokens/currency/trait.Currency.html#tymethod.free_balance) and [`transfer()`](https://substrate.dev/rustdocs/latest/frame_support/traits/tokens/currency/trait.Currency.html#tymethod.transfer) to another account.

~~I haven't rewritten the test cases yet, but want to get your feedback first on what you think about the current implementation. We can do the rewrite on the tutorial later as well.~~ 
Test cases are done now.

### DIFF
--- a/pallets/kitties/src/lib.rs
+++ b/pallets/kitties/src/lib.rs
@@ -15,6 +15,7 @@ pub mod pallet {
 	use frame_support::{
 		sp_runtime::traits::Hash,
 		traits::{ Randomness, Currency, tokens::ExistenceRequirement },
+		transactional
 	};
 	use sp_io::hashing::blake2_128;
 
@@ -76,8 +77,8 @@ pub mod pallet {
 		KittyNotExist,
 		/// Handles checking that the Kitty is owned by the account transferring, buying or setting a price for it.
 		NotKittyOwner,
-		/// Ensurers the Kitty is for sale.
-		KittyNotOnSale,
+		/// Ensures the Kitty is for sale.
+		KittyNotForSale,
 		/// Ensures that the buying price is greater than the current price.
 		KittyBidPriceTooLow,
 		/// Ensures that an account has enough funds to purchase a Kitty. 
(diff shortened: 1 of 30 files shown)


pallets/kitties/src/lib.rs
```diff
@@ -31,99 +36,68 @@ pub mod pallet {
...
 
 	// Errors.
 	#[pallet::error]
 	pub enum Error<T> {
-		/// Nonce has overflowed past u64 limits
-		NonceOverflow,
+		KittyCntOverflow,
+		ExceedMaxKittyOwned,
+		KittyNotExist,
+		KittyNotFound,
```
sacha-l: Can't we also use `KittyNotExist` instead of `KittyNotFound` here ?

pallets/kitties/src/lib.rs
```diff
@@ -379,6 +370,9 @@ pub mod pallet {
 
 			// Update the kitty owner
 			kitty.owner = to.clone();
+			if reset_price {
+				kitty.price = None;
```
sacha-l, line 362: This relates to the logic in `buy_kitty` which shouldn't reset the price to `None` but really should reset it to `bid_price`.

pallets/kitties/src/lib.rs
```diff
@@ -127,6 +127,13 @@ pub mod pallet {
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
+		#[pallet::weight(1)]
+		pub fn hello_world(origin: OriginFor<T>) -> DispatchResult {
+			let sender = ensure_signed(origin)?;
+			log::info!("Hello World! Transaction received from: {}", sender);
+			Ok(())
+		}
```
sacha-l, line 135: For some reason this kept throwing an error:

```
  |
  149 |             log::info!("Hello World! Transaction received from: {}", sender);
      |                                                                      ^^^^^^ `<T as frame_system::Config>::AccountId` cannot be formatted with the default formatter
      |
      = help: the trait `Display` is not implemented for `<T as frame_system::Config>::AccountId`
```

pallets/kitties/src/lib.rs
jimmychu0807: Changed this to `KittyNotExist`

pallets/kitties/src/lib.rs
jimmychu0807: I want to demonstrate how to log a message to the console here. In my latest commit, I have merged this in `create_kitty()` function and removed this `hello_world()` method. 

To compile this without error, the code should update to:

```rust
log::info!("Hello World! Transaction received from: {:?}", sender);
```

pallets/kitties/src/lib.rs
jimmychu0807: Having the price field set means the kitty is for sale and it can be sold in the next instance by anyone willing to pay for this price. I assume when a buyer purchase a kitty, he may not want to sell it immediately, so setting the price to `None`. When he is ready to sell the kitty, he will first call `set_price()`, setting the ask price and putting the kitty for sale. 

Same thinking goes for `transfer()`.

jimmychu0807:
@sacha-l I have updated the main kitty code, comments, and the testing code (all tests pass now). I think the code is good for review now.

pallets/kitties/src/lib.rs
sacha-l: Makes sense! Let's leave it.

PR #235, opened 2021-09-07
Remove now irrelevant workflow

jeluard:


NukeManDan:
So we don't use this at all now? Might be useful for #236 ?

jeluard:
@NukeManDan This is solely for playground, you can trust me ;)

NukeManDan (approved):

PR #273, opened 2021-12-01
Upgrade to polkadot-v0.9.12

### LINKED ISSUE #272: Upgrade template to polkadot-v09.12
Upgrade template to use substrate branch `polkadot-v0.9.12`

al3mart:
closes #272

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,5 +6,4 @@ members = [
     'node',
     'pallets/template',
     'runtime',
-    '',
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,10 @@
+[profile.release]
+panic = 'unwind'
+
 [workspace]
 members = [
     'node',
     'pallets/template',
     'runtime',
+    '',
 ]
-[profile.release]
-panic = 'unwind'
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -1,8 +1,10 @@
+[profile.release]
+panic = 'unwind'
+
 [workspace]
 members = [
     'node',
     'pallets/template',
     'runtime',
+    '',
```
NukeManDan, line 9:
```suggestion
```

jimmychu0807:
Hi @al3mart,

I took a look at what you did, and I don't have the context about this PR. So I am throwing what I know:

- Usually the `main` branch of node-template is following the [`monthly` tag release from Substrate](https://github.com/paritytech/substrate/tree/monthly-2021-12).
- They are usually quite different from [Polkadot releases](https://github.com/paritytech/polkadot/tree/v0.9.14)

So, if we do want to maintain two branches, one for Substrate release and another for Polkadot release, we should have a consensus internally which two branches are going to dedicate for doing that. And these two branches very probably will never merge together. 

But I do agree there maybe a need to maintain a separate branch of node-template for following polkadot release. This branch will make people developing parachain easier.

al3mart:
Right, then there is no need of merging this, it could stay as a separate branch :)

PR #285, opened 2021-12-21
README: Persistent state fix

### LINKED ISSUE #284: README incorrectly describes how to run with persistent state
@NukeManDan 

**Motivation**

The README has the following incorrect guidance:

image

But, this doesn't work. There will be no state persisted between runs.

**Suggested Solution**

Update this bit of the README to describe how to run the template with persisted state; maybe mention what the `dev` flag does to give the commands more color.

al3mart:
Closes #284

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ node.
 
 ### Single-Node Development Chain
 
-This command will start the single-node development chain with persistent state:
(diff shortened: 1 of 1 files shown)


NukeManDan (approved):
Great!! @al3mart can you also upstream this to Substrate?

PR #294, opened 2022-01-04
Incorporate ConstU32

pawanbisht62:
Incorporated ConstU32 and removed paramater_types.


NukeManDan:
If you think this is needed, please instead detail why in a PR upstream https://github.com/paritytech/substrate/tree/master/bin/node-template

Issue #296, opened 2022-01-18
Failed to build the release with 'edition2021' is required 


James-Sangalli:
Using commit: `b74af51`

Current setup: 

> Default host: x86_64-apple-darwin
rustup home:  /Users/jamessangalli/.rustup

installed toolchains
--------------------

stable-x86_64-apple-darwin (default)
nightly-x86_64-apple-darwin
1.51.0-x86_64-apple-darwin

installed targets for active toolchain
--------------------------------------

wasm32-unknown-unknown
x86_64-apple-darwin

active toolchain
----------------

stable-x86_64-apple-darwin (default)
rustc 1.58.0 (02072b482 2022-01-11)

Error: 

> error: failed to load manifest for workspace member `/Users/jamessangalli/Documents/projects/substrate-node-template/node`

> Caused by:
  failed to parse manifest at `/Users/jamessangalli/Documents/projects/substrate-node-template/node/Cargo.toml`

> Caused by:
  feature `edition2021` is required

 > consider adding `cargo-features = ["edition2021"]` to the manifest

Anyone encountered this before? Should I downgrade my rustc to a specific version? 

Thank you

NukeManDan:
Good catch! https://github.com/paritytech/substrate/pull/10710 updates this, somehow missed here.

PR #304, opened 2022-01-28
fix: add edition 2021 to manifest #296

James-Sangalli:
The fix proposed for #296 did not work for me, but this one does.


NukeManDan:
Strange: this is not needed upstream at all, and no other project I know of need this. Perhaps you missed the fix to get 2021 in the edition field?

PR #312, opened 2022-02-23
add multisig pallet configuration in runtime

ahmedali8:


NukeManDan:
Thanks! But the template won't add this pallet, if you see an explicit reason to do so, please add this upstream and make a case to the substrate maintainers as to why :pray:

ahmedali8:
> Thanks! But the template won't add this pallet, if you see an explicit reason to do so, please add this upstream and make a case to the substrate maintainers as to why 🙏

No problem, Thank you so much for checking out the PR.

PR #323, opened 2022-03-11
Improvements to Kitties tests

sacha-l:
Improvements include:
- Comments to add clarity to code
- Adding testing logic for transferring balances 
- Fixed a bug in the pallet caught by `transfer_kitty_should_work`: bid price never got debited correctly so seller would just get their original price from sale, and not the actual bid price.
- Refactored a whole bunch, trying to keep to the "works / fails" pattern for each dispatchable (down from 21 tests to 11).

### DIFF
--- a/pallets/kitties/src/lib.rs
+++ b/pallets/kitties/src/lib.rs
@@ -367,16 +367,17 @@ pub mod pallet {
 
 			// Mutating state here via a balance transfer, so nothing is allowed to fail after this.
 			if let Some(bid_price) = maybe_bid_price {
+				// Current kitty price 
 				if let Some(price) = kitty.price {
 					ensure!(bid_price >= price, Error::<T>::BidPriceTooLow);
 					// Transfer the amount from buyer to seller
-					T::Currency::transfer(&to, &from, price, ExistenceRequirement::KeepAlive)?;
+					T::Currency::transfer(&to, &from, bid_price, ExistenceRequirement::KeepAlive)?;
 					// Deposit sold event
 					Self::deposit_event(Event::Sold {
 						seller: from.clone(),
 						buyer: to.clone(),
 						kitty: kitty_id,
-						price,
+						price: bid_price,
 					});
 				} else {
 					return Err(Error::<T>::NotForSale.into())
--- a/pallets/kitties/src/lib.rs
+++ b/pallets/kitties/src/lib.rs
@@ -265,7 +265,7 @@ pub mod pallet {
 
 	impl<T: Config> Pallet<T> {
 		// Generates and returns DNA and Gender
-		fn gen_dna() -> ([u8; 16], Gender) {
+		pub fn gen_dna() -> ([u8; 16], Gender) {
 			// Create randomness
 			let random = T::KittyRandomness::random(&b"dna"[..]).0;
 
@@ -283,31 +283,41 @@ pub mod pallet {
 
 			// Generate Gender
 			if hash[0] % 2 == 0 {
+				// Males are identified by having a even leading byte
 				(hash, Gender::Male)
 			} else {
+            	// Females are identified by having a odd leading byte
 				(hash, Gender::Female)
 			}
 		}
 
 		// Picks from existing DNA
 		fn mutate_dna_fragment(dna_fragment1: u8, dna_fragment2: u8, random_value: u8) -> u8 {
+			// Given some random u8 
 			if random_value % 2 == 0 {
+				// either return `dna_fragment1` if its an even value 
 				dna_fragment1
 			} else {
+				// or return `dna_fragment2` if its an odd value
 				dna_fragment2
 			}
 		}
 
 		// Generates a new kitty using existing kitties
 		pub fn breed_dna(parent1: &[u8; 16], parent2: &[u8; 16]) -> ([u8; 16], Gender) {
+			
+			// Call `gen_dna` to generate random kitty DNA 
+			// We don't know what Gender this kitty is yet
 			let (mut new_dna, new_gender) = Self::gen_dna();
 
+			// randomly combine DNA using `mutate_dna_fragment`
 			for i in 0..new_dna.len() {
 				// At this point, `new_dna` is a randomly generated set of bytes, so we can
-				// extract each of its bytes to act as a random value.
+				// extract each of its bytes to act as a random value for `mutate_dna_fragment`
 				new_dna[i] = Self::mutate_dna_fragment(parent1[i], parent2[i], new_dna[i])
 			}
-			(new_dna, new_gender)
+			// return new DNA and gender
+			(new_dna, new_gender)	
 		}
 
 		// Helper to mint a kitty
--- a/pallets/kitties/src/tests.rs
+++ b/pallets/kitties/src/tests.rs
@@ -174,10 +174,7 @@ fn breed_kitty_works() {
 		// Breeder can only breed kitties they own
 		assert_ok!(SubstrateKitties::breed_kitty(Origin::signed(1), mom, dad));
 
-		// Check that newly bred kitty exists
-		assert_ok!(KittiesOwned::<Test>::get(1)[3]);
-
-		// Check the new DNA is from the mom and dad
+		// Check the new kitty exists and DNA is from the mom and dad
 		let new_dna = KittiesOwned::<Test>::get(1)[3];
 		for &i in new_dna.iter() {
 			assert!(i == 0u8 || i == 1u8)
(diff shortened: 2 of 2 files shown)


pallets/kitties/src/lib.rs
```diff
@@ -367,16 +377,17 @@ pub mod pallet {
 
 			// Mutating state here via a balance transfer, so nothing is allowed to fail after this.
 			if let Some(bid_price) = maybe_bid_price {
+				// Current kitty price 
 				if let Some(price) = kitty.price {
 					ensure!(bid_price >= price, Error::<T>::BidPriceTooLow);
 					// Transfer the amount from buyer to seller
-					T::Currency::transfer(&to, &from, price, ExistenceRequirement::KeepAlive)?;
+					T::Currency::transfer(&to, &from, bid_price, ExistenceRequirement::KeepAlive)?;
```
shawntabrizi, line 384: this seems like a wrong change. We want to charge the user for the actual price of the kitty, not the `bid_price` which may be an overestimate.

sacha-l: Wouldn't this result in `buy_kitty` ignoring the `bid_price` each time it's called with a higher price than what's set?  i.e. the sale would always use the actual price of the kitty and not the proposed (higher) price.

sacha-l: Resolved in https://github.com/substrate-developer-hub/substrate-node-template/pull/323/commits/4f1110ae7ddcfc00aa1c35f03bcfdfc6183a5af4.

pallets/kitties/src/lib.rs
```diff
@@ -283,31 +283,41 @@ pub mod pallet {
 
 			// Generate Gender
 			if hash[0] % 2 == 0 {
+				// Males are identified by having a even leading byte
 				(hash, Gender::Male)
 			} else {
+            	// Females are identified by having a odd leading byte
```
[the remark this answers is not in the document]
sacha-l: Resolved in https://github.com/substrate-developer-hub/substrate-node-template/pull/323/commits/4f1110ae7ddcfc00aa1c35f03bcfdfc6183a5af4.

pallets/kitties/src/tests.rs
```diff
@@ -345,90 +299,46 @@ fn buy_kitty_works() {
...
-		let id = KittiesOwned::<Test>::get(2)[0];
-		let price = u64::MAX;
-		assert_ok!(SubstrateKitties::set_price(Origin::signed(2), id, Some(price)));
-
+		// Account 3 can't buy with insufficient funds
 		assert_noop!(
-			SubstrateKitties::buy_kitty(Origin::signed(1), id, price),
+			SubstrateKitties::buy_kitty(Origin::signed(3), id, u64::MAX),
 			pallet_balances::Error::<Test>::InsufficientBalance
 		);
```
sacha-l, line 336: I updated the `assert_noop!` with a fresh account here. But I noticed that if we use account 2, the test fails with: 
```bash
---- tests::buy_kitty_works stdout ----
thread 'tests::buy_kitty_works' panicked at 'assertion failed: `(left == right)`
  left: `Ok(())`,
 right: `Err(Module { index: 1, error: 2, message: Some("InsufficientBalance") })`', pallets/kitties/src/tests.rs:333:9
```
Why is this?

pallets/kitties/src/tests.rs
```diff
@@ -84,69 +117,85 @@ fn transfer_kitty_should_work() {
...
+		// Account #1 bids to buy kitty at 2x the price
+		assert_ok!(SubstrateKitties::buy_kitty(Origin::signed(1), id, set_price * 2));
+
+		// Check that balance transfer works as expected
+		let balance_1_after = Balances::free_balance(&1);
+		let balance_2_after = Balances::free_balance(&2);
+
+		// The amount transferred should be the original kitty price 
+		assert!(balance_1_before - set_price == balance_1_after);
+		assert!(balance_2_before + set_price == balance_2_after);
```
shawntabrizi, line 153: this seems like it should be a different test. This test is labeled for testing `transfer`

pallets/kitties/src/tests.rs
```diff
@@ -84,69 +117,85 @@ fn transfer_kitty_should_work() {
...
+		let balance_1_before = Balances::free_balance(&1);
+		let balance_2_before = Balances::free_balance(&2);
+
+		// Account #2 sets new price to 4
+		let id = KittiesOwned::<Test>::get(2)[0];
+		let set_price = 4;
+		assert_ok!(SubstrateKitties::set_price(Origin::signed(2), id, Some(set_price)));
+
+		// Account #1 bids to buy kitty at 2x the price
+		assert_ok!(SubstrateKitties::buy_kitty(Origin::signed(1), id, set_price * 2));
```
shawntabrizi, line 145: this is not an accurate comment

shawntabrizi: please update `bid_price` to `limit_price`

pallets/kitties/src/tests.rs
```diff
@@ -263,76 +257,36 @@ fn dna_helpers_should_work() {
...
-fn transfer_fails_when_no_kitty_exists() {
-	new_test_ext(vec![
-		(1, *b"1234567890123456", Gender::Female),
-		(2, *b"123456789012345a", Gender::Male),
-	])
-	.execute_with(|| {
-		// Check transfer fails when no kitty exists
-		let random_id = [0u8; 16];
+		// Random values should not be equal
+		assert_ne!(random_dna_1, random_dna_2);
```
shawntabrizi, line 276: this test seems out of place. The test is named "breed kitty fails"

pallets/kitties/src/tests.rs
```diff
@@ -263,76 +257,36 @@ fn dna_helpers_should_work() {
...
+		assert_ok!(SubstrateKitties::mint(&3, kitty_1, Gender::Female));
+		assert_ok!(SubstrateKitties::mint(&3, kitty_2, Gender::Female));
 
-		// Account #10 should not be able to receive a new kitty
+		// Same gender kitty can't breed
 		assert_noop!(
-			SubstrateKitties::transfer(Origin::signed(1), 10, kitty_1),
-			Error::<Test>::TooManyOwned
+			SubstrateKitties::breed_kitty(Origin::signed(3), kitty_1, kitty_2),
+			Error::<Test>::CantBreed
```
shawntabrizi, line 289: Check that breed kitty fails with too many kitties

pallets/kitties/src/tests.rs
```diff
@@ -342,93 +296,53 @@ fn buy_kitty_works() {
...
+		let balance_2_before = Balances::free_balance(&2);
 
+		// Account #2 sets a price of 4 for their kitty
 		assert_ok!(SubstrateKitties::set_price(Origin::signed(2), id, Some(set_price)));
 
-		// Account #1 can buy account #2's kitty
-		assert_ok!(SubstrateKitties::buy_kitty(Origin::signed(1), id, set_price));
+		// Account #1 can buy account #2's kitty at some bid_price
+		let bid_price = 6;
+		assert_ok!(SubstrateKitties::buy_kitty(Origin::signed(1), id, bid_price));
```
shawntabrizi, line 313: this test seems duplicated from above. probably should be removed above

pallets/kitties/src/tests.rs
```diff
@@ -342,93 +296,53 @@ fn buy_kitty_works() {
...
 
-		// Add the fact the price transfer worked
+		// Check balance transfer works as expected
+		let balance_1_after = Balances::free_balance(&1);
+		let balance_2_after = Balances::free_balance(&2);
 
+		// We use set_price as this is the amount actually being charged
+		assert!(balance_1_before - set_price == balance_1_after);
+		assert!(balance_2_before + set_price == balance_2_after);
```
shawntabrizi, line 321: there is `assert_eq!(a, b)` which is more ergonomic and gives better errors

pallets/kitties/src/tests.rs
```diff
@@ -342,93 +296,53 @@ fn buy_kitty_works() {
...
 		// Check buy_kitty fails when balance is too low
+		// Get the balance of account 3
+		let balance_3 = Balances::free_balance(&3);
 
-		// Use some kitty in storage owned by account 2 and set a high price
-		let id = KittiesOwned::<Test>::get(2)[0];
-		let price = u64::MAX;
-		assert_ok!(SubstrateKitties::set_price(Origin::signed(2), id, Some(price)));
+		// Reset the price to something higher than account 3's balance
+		assert_ok!(SubstrateKitties::set_price(Origin::signed(1), id, Some(balance_3*10)));
```
shawntabrizi, line 334: Why is this test inside "buy_kitty_works"?

shawntabrizi (approved):

PR #347, opened 2022-05-20
Runtime without Frame

JoshOrndorff:
This is a revival of a project I started a long time ago but never finished. (https://github.com/JoshOrndorff/recipes/pull/144)

This demonstrates a simple runtime written entirely without FRAME. It directly implements the necessary runtime APIs. The Runtime's logic is quite simple. It just maintains a single bit of state. The bit can be set, cleared, or toggled by anyone. Transactions require the user (or their wallet) to provide a salt to de-duplicate transactions.

```bash
# Run the node in the usual way
node-template --dev`


# In another terminal, check that the state starts cleared. (result: 0x00)
curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{
       "jsonrpc":"2.0",
        "id":1,
        "method":"state_getStorage",
        "params": ["0x626F6F6C65616E"]
      }'
{"jsonrpc":"2.0","result":"0x00","id":1}

# Submit a Toggle transaction with salt 3 signed by alice
curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{
       "jsonrpc":"2.0",
        "id":1,
        "method":"author_submitExtrinsic",
        "params": ["0x02032a51e39e68e0736b93a80d84d6d9fcb77f60d0d87e158e48191e3a00e1aa647d303b50c1fc15a89650a84af6273635697888fba039b53a38500935d6863d5c82d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"]
      }'
{"jsonrpc":"2.0","result":"0xa9b156e072f43817c923de3184a6f2934558ec3cdf81ec07b5160dca63633b2a","id":1}

# Check the state again and notice that the state has been set. (result: 0x01)
curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{
       "jsonrpc":"2.0",
        "id":1,
        "method":"state_getStorage",
        "params": ["0x626F6F6C65616E"]                                                                                                                                                                                      
      }'
{"jsonrpc":"2.0","result":"0x01","id":1}
```


JoshOrndorff:
Woah! I think it actually included a transaction!

```bash
# Submit the transaction `0x0004` via RPC
$ curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{
       "jsonrpc":"2.0",
        "id":1,
        "method":"author_submitExtrinsic",
        "params": ["0x0004"]
      }'
{"jsonrpc":"2.0","result":"0x7324cf9181f377b38e2edeca7a080796554246690656566def41afedef911d0d","id":1}

# Node output in other terminal
2022-05-20 15:56:57 ✨ Imported #63 (0xd48c…7181)    
2022-05-20 15:57:00 🙌 Starting consensus session on top of parent 0xd48c750a262f4acc203a844cd5af5ecf97066de782053e1a4476af74b4b27181    
2022-05-20 15:57:00 🎁 Prepared block for proposing at 64 (0 ms) [hash: 0xba15a9830087d539ca0681bd229f9c6a6053df3117e7417eb042d2928ebdb442; parent_hash: 0xd48c…7181; extrinsics (1): [0x7324…1d0d]]    
2022-05-20 15:57:00 🔖 Pre-sealed block for proposal at 64. Hash now 0x204d979504f6eb866a00ec5630e544dcae7e44ec1c4008034935b4bb9d2db467, previously 0xba15a9830087d539ca0681bd229f9c6a6053df3117e7417eb042d2928ebdb442. 
```

I'm pretty sure that never worked before! cc @kianenigma

Next step is to see if the state actually updated.

JoshOrndorff:
Wow, indeed the storage is updated! I can hardly believe it.

```bash
curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{
       "jsonrpc":"2.0",
        "id":1,
        "method":"state_getStorage",
        "params": ["0x626F6F6C65616E"]
      }'
{"jsonrpc":"2.0","result":"0x01","id":1}
```

Next step: better provides tags or some kind of nonce or salt so transactions don't get temporarily banned.

Also, I think my set and clear variants might be swapped or something.

frameless-runtime/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+/// of data like extrinsics, allowing for them to continue syncing the network through upgrades
+/// to even the core datastructures.
+pub mod opaque {
+	use super::*;
+
+	/// Opaque block header type.
+	pub type Header = generic::Header<BlockNumber, BlakeTwo256>;
+	/// Opaque block type.
+	pub type Block = generic::Block<Header, FramelessTransaction>;
+
```
kianenigma, line 74:
```suggestion
```

these seem unused.

frameless-runtime/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+/// The Extrinsic type for this runtime. Currently extrinsics are unsigned.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize, parity_util_mem::MallocSizeOf))]
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
+pub enum FramelessCall {
+	Set,
+	Clear,
+	Toggle,
+}
+
+// So we can have a cli utility to create transactions. Not necessary for the runtime itself.
```
kianenigma, line 161: In FRAME, we solve this with metadata ;)

frameless-runtime/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+		}
+	}
+}
+
+/// Checks a transaction that came in from the outside world.
+/// Returns the same type needed by the transaction pool.
+/// This is seperated into a function so it can be re-used
+/// when executing transactions.
+fn check_framesless_transaction(tx: &FramelessTransaction) -> TransactionValidity {
+
```
kianenigma, line 202:
```suggestion
```

frameless-runtime/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+			}
+
+			// In frame executive, they call final_checks, but that might be different
+			Self::finalize_block();
+		}
+
+		fn initialize_block(header: &<Block as BlockT>::Header) {
+			info!(target: "frameless", "🖼️ Entering initialize_block. header: {:?}", header);
+			// Store the header info we're given for later use when finalizing block.
+			sp_io::storage::set(&HEADER_KEY, &header.encode());
```
kianenigma, line 260: can we have a runtime that just ignores this?

frameless-runtime/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+			info!(target: "frameless", "🖼️ Newly stored state is: {:?}", next_state);
+
+			sp_io::storage::set(&BOOLEAN_KEY, &next_state.encode());
+			Ok(Ok(()))
+		}
+
+		fn finalize_block() -> <Block as BlockT>::Header {
+			info!(target: "frameless", "🖼️ Entering finalize block.");
+			// https://substrate.dev/rustdocs/master/sp_runtime/generic/struct.Header.html
+			let raw_header = sp_io::storage::get(&HEADER_KEY)
```
kianenigma, line 295: yeah, not sure why we write the header?

kianenigma: ahh `finalize_block` expects it, I see.

frameless-runtime/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+			tx: <Block as BlockT>::Extrinsic,
+			block_hash: <Block as BlockT>::Hash,
+		) -> TransactionValidity {
+			info!(target: "frameless", "🖼️ Entering validate_transaction. source: {:?}, tx: {:?}, block hash: {:?}", source, tx, block_hash);
+
+			check_framesless_transaction(&tx)
+		}
+	}
+
+	impl sp_api::Metadata<Block> for Runtime {
```
kianenigma, line 338: can't this be skipped? I think the client can also be tweaked to not support it.

node/src/cli.rs
```diff
@@ -47,4 +47,19 @@ pub enum Subcommand {
...
+pub struct CreateTransactionCmd {
+	/// The action to take: "set", "clear", or "toggle"
+	pub action: node_template_runtime::FramelessCall,
+
+	/// The salt to use in the transaction. If none is supplied, a "random" one will be chosen
+	pub salt: Option<u8>,
+
+	/// The seed to sign with. If none is provided, Alice's will be used.
+	pub seed: Option<String>
+}
```
kianenigma, line 65:
```suggestion
}

```

frameless-runtime/src/lib.rs
JoshOrndorff: The runtime can compile without this, but it cannot be used by a Substrate client.

I asked this previously as well. Perhaps a PR to Substrate makes sense?

https://github.com/paritytech/substrate/blob/master/client/service/src/builder.rs#L399-L400

frameless-runtime/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+		fn metadata() -> OpaqueMetadata {
+			OpaqueMetadata::new(vec![0])
+		}
+	}
+
+	impl sp_offchain::OffchainWorkerApi<Block> for Runtime {
```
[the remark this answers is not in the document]
JoshOrndorff: Same. https://github.com/paritytech/substrate/blob/master/client/service/src/builder.rs#L399-L400

frameless-runtime/src/lib.rs
JoshOrndorff: No, we need to return a header when we finalize the block. The simplest thing I could think to do (which I've done here) is to store the one we're given here. Then when finalizing, update the state root, and return it with no other modifications.

frameless-runtime/src/lib.rs
JoshOrndorff: Is Metadata  frame-specific concept? It kind of seems like it is to me. If so, that's another reason that the metadata api shouldn't be required.

frameless-runtime/src/lib.rs
JoshOrndorff: They are used on the client side. `use frameless_runtime::opaque::Block` in `service.rs`. So simply deleting them makes the node not compile.

But this is something I haven't really thought through. Should this type be in a primitives crate or something? I just have it here because it is copied from the node template.

frameless-runtime/src/lib.rs
JoshOrndorff: https://github.com/paritytech/substrate/issues/11743

I've asked whether this runtime api should really be mandatory.

frameless-runtime/src/lib.rs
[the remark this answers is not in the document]
JoshOrndorff: https://github.com/paritytech/substrate/issues/11742

I've asked whether this api should be required. I think it's pretty reasonable to expect that this one should not be required.

frameless-runtime/src/lib.rs
```diff
@@ -0,0 +1,412 @@
...
+			let raw_header = sp_io::storage::get(&HEADER_KEY)
+				.expect("We initialized with header, it never got mutated, qed");
+
+			// Clear the raw header out of storage when we are done with it.
+			sp_io::storage::clear(&HEADER_KEY);
+
+			let mut header = <Block as BlockT>::Header::decode(&mut &*raw_header)
+				.expect("we put a valid header in in the first place, qed");
+
+			let raw_state_root = &sp_io::storage::root(sp_storage::StateVersion::default())[..];
```
JoshOrndorff, line 303: @kianenigma Currently this runtime does not modify the header that was passed in at all except for the state root. I'm starting to wonder if we need to recalculate the extrinsic root in the runtime. This is what [FRAME does](https://github.com/paritytech/substrate/blob/ee3eb8f2448cc1bb978c5d1564febd351c128bb0/frame/system/src/lib.rs#L1350).

JoshOrndorff: Indeed we do need to calculate the extrinsic root. TODO

JoshOrndorff:
So I just noticed something weird here. I can start two nodes, and they will peer together. But the moment that one of the authors a block, the un-peer. No interesting logs from the non-authoring node. I would expect it to at least print the log that it was in the execute block function. But even that doesn't happen. I'm pretty sure I tested that before, but I don't see any notes about it above. :-/

I'm not sure how to debug that.

NukeManDan:
@JoshOrndorff I think you had work for the https://polkadot.network/academy/ related to a substrate based frameless runtime that was a stand-alone repo... is that related to this work and also public? Can't find it if so, only in the private (planned public first half this year for sure) content repo. https://github.com/Polkadot-Blockchain-Academy/pba-content/tree/main/syllabus/4-Substrate/3-Runtime_and_Host_Functions/frameless-node-template

Asking as if work is concentrated in that priv repo, can this PR be closed? Or are they separate things?

JoshOrndorff:
I don't have this in a standalone repo. Someone copied it from here into the pba content repo. You may close this PR.

PR #357, opened 2022-07-19
Fix docker-compose.yml

alvicsam:
`docker-compose.yml` uses an outdated image. The PR fixes it.

cc https://github.com/paritytech/substrate/issues/11750


docker-compose.yml
```diff
@@ -3,7 +3,7 @@ version: "3.2"
 services:
   dev:
     container_name: node-template
-    image: paritytech/ci-linux:974ba3ac-20201006
+    image: paritytech/ci-linux:production
```
NukeManDan, line 6: @alvicsam won't this break over time for old tags? It _should_ mostly be fine... but perhaps it's best to update this on releases to a fixed known image?

alvicsam: I'm not sure honestly, the main purpose of this PR was to fix it for current main. With `974ba3ac-20201006` version the `docker-compose.yml` didn't work for `main` at all

PR #369, opened 2022-09-01
improve(node): update clone client to avoid borrowed

pidb:
In this description of [Substrate tutorials](https://docs.substrate.io/tutorials/work-with-pallets/contracts-pallet/#expose-the-contracts-api)

```rust
module.merge(System::new(client.clone(), pool.clone(), deny_unsafe).into_rpc())?;
module.merge(TransactionPayment::new(client.clone()).into_rpc())?;
module.merge(Contracts::new(client.clone()).into_rpc())?; // Add this line
```

However, `client` does not actually have clone, and there will got `value borrowed here after move` error when compiling.

https://github.com/substrate-developer-hub/substrate-node-template/blob/6a8b2b12371395979099d2c79ccc1860531b0449/node/src/rpc.rs#L49

So, I want to add `client.clone()` to avoid this error :)

### DIFF
--- a/node/src/rpc.rs
+++ b/node/src/rpc.rs
@@ -46,7 +46,7 @@ where
 	let FullDeps { client, pool, deny_unsafe } = deps;
 
 	module.merge(System::new(client.clone(), pool.clone(), deny_unsafe).into_rpc())?;
-	module.merge(TransactionPayment::new(client).into_rpc())?;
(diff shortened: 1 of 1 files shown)


NukeManDan:
@pidb please do make this upstream as this template isn't the place to make changes, it's only purpose is to be a stand alone copy over from upstream releases. :pray: https://github.com/paritytech/substrate/tree/master/bin/node-template

PR #382, opened 2022-10-24
Update to latest enum syntax

sacha-l:
Updates `SomethingStored(u32, T::AccountId),` to `SomethingStored { something: u32, who: T::AccountId },` to follow latest FRAME syntax.


NukeManDan:
This should be moved upstream, this template is not the "real source"

PR #395, opened 2022-11-15
Remove unused source command_helper.rs

h4n0:
As command_helper.rs has already been removed in [main.rs]( https://github.com/substrate-developer-hub/substrate-node-template/blob/main/node/src/main.rs) for a while, and the code in this source file has not been maintained since April, I suggest removing this file to avoid any confusion and keep a clean source list.


al3mart (approved):

al3mart:
Thanks!

PR #410, opened 2023-01-18
Remove all quotes from the manifest key  in all Cargo.toml files

jacob-chia:


lisa-parity (approved):

lisa-parity:
@coderobe, can we get this updated to use without quotes around "branch" like:
branch = "${polkadot_branch}"

and I'm wondering why the `substrate-node-template` has fallen behind and has `Cargo.toml` files referencing `0.9.35` instead of 36 or 37. I think we should be at 37, no?

coderobe:
what is the purpose of this?

coderobe:
this should probably be an upstream issue in https://github.com/paritytech/diener

jacob-chia:
> 
Well, the style of  `"branch" = ...`  works, it just comes with a warning as below:
(IDE: Vscode + Better TOML extension)
image

And I tried the `diener` you mentioned. When I did `diener update --all --branch polkadot-v0.9.36`, the quotes of the `branch` were gone.
image

Since the warining will be fixed the next time `diener update`, this PR can be closed.

PR #416, opened 2023-03-07
Undeploy insecure randomness pallet

ggwpez:
The `randomness-collective-flip` pallet was [recently renamed](https://github.com/paritytech/substrate/pull/13301) to `insecure-randomness-collective-flip`.  
As the name says; it is insecure should not be used in production. Therefore undeploying it from this template to not have people accidentally deploy it.


ggwpez:
> It's used in some of the tutorials AFAIK so might be worth keeping here.

Why is it used in tutorials? The pallet should never be used in production.  
Teaching about it in a tutorial might give of the idea that it should be used…

NukeManDan:
This template always matches upstream on Polkadot releases, I see https://github.com/paritytech/substrate/pull/13560 has this diff so it will make it's way here shortly too. I don't see a real hard requirement in getting this in _now_ seeing as where it's referenced in docs already points to some older version of the template anyhow.

IMHO any use of the pallet in docs should be fast tracked to deprecation.

ggwpez:
>This template always matches upstream on Polkadot releases, I see https://github.com/paritytech/substrate/pull/13560 has this diff so it will make it's way here shortly too

Ah okay - I dont know how it works :sweat_smile:. So i can close this?

NukeManDan:
Yeah, up to the release team to help push this to us upstream now :grin:

PR #519, opened 2024-02-13
run cargo update ahash@0.7.6

thiolliere:
ahash version 0.7.6 no longer works with latest version of rustc nightly because `stdsimd` features has been removed.

To fix it I updated ahash with the command:
```
 cargo update ahash@0.7.6
```

This should fix the compilation https://github.com/substrate-developer-hub/substrate-docs/issues/2130

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,9 +149,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.7.6"
+version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


thiolliere:
will be fixed by https://github.com/paritytech/polkadot-sdk/pull/3296

bkchr:
You would still need to update the `Cargo.lock` here as I'm doing also the same in the SDK repo.

thiolliere:
Yes I updated ahash 0.8 (and also syn because cargo asked to)
I put resolver 2 similarly to your PR

bkchr (approved):

PR #523, opened 2024-07-17
Fix nix devshell and provides rust based on rust-toolchain.toml

yanganto:
The previous Nix dev shell is improper and incomplete, so it will not always work as a reproducible shell.  

This PR provides a complete rust toolchain based on `rust-toolchain.toml`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,12 +1,15 @@
 {
   "nodes": {
     "flake-utils": {
+      "inputs": {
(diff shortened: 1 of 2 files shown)


Gesge:
What's up guy?This is my automatic reply~ With the purpose of enjoying the holiday,I did this auto-reply~Please wait a while but no too long.As soon as I saw it,turn back to your letter~~~

